PR #13, opened 2018-05-23
fix concurrency broken maps

Geosearchef:


Geosearchef:
tested to be deterministic

n3roxis (approved):

PR #23, opened 2019-07-09
Feature/cli args parsing, --with-biome arg

Geosearchef:


src/java/generator/MapGenerator.java
```diff
@@ -190,9 +246,9 @@ public SCMap generate(long seed) throws ExecutionException, InterruptedException
 
 		map.setTextureMaskLow(grassTexture.getFloatMask(), lightGrassTexture, rockTexture.getFloatMask(), new FloatMask(513, 0));
 
-		Biome biomeSet = Biomes.list.get(random.nextInt(Biomes.list.size()));
+		Biome biomeSet = BIOME.orElseGet(() -> Biomes.getRandomBiome(random));
 
-		System.out.printf(String.format("Using biome %s", biomeSet.getName()));
+		System.out.printf("Using biome %s\n", biomeSet.getName());
```
louvenarde, line 196: This is a regression, please undo :P 
Biome name is supposed to be on the same line than the sentence it's part of 
"Using biome Refuge"
String format can be removed because printf exists, but the newline is unnecessary

Geosearchef: The newline is after the biome name

louvenarde: nvm then! ready for me

Rackover (approved):

Rackover:
(Untested from me)

n3roxis (commented):
Did anyone test this?

Rackover:
Yes I tested it and it works as intended, as far as my testing goes

PR #31, opened 2020-06-15
Visual debugger

Katharsas:
Can somebody please test this a bit or something.
Usage:
Set `VisualDebugger.ENABLED` to true, then to draw mask operations for all masks `VisualDebugger.startRecordAll()` or to only draw operations of specific masks `mask.startVisualDebugger()`. Use debugger to step through operations to see each step clearly or run in normal mode for realtime visualization.

When `VisualDebugger.ENABLED` is true, concurrent masks are reverted to be synchronous to make proper stepping through easy, this reduces performance obviously.

### DIFF
--- a/src/java/generator/VisualDebugger.java
+++ b/src/java/generator/VisualDebugger.java
@@ -4,9 +4,7 @@ import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.awt.image.ColorModel;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Map;
-import java.util.Set;
 
 import map.BinaryMask;
 import map.FloatMask;
@@ -14,11 +12,14 @@ import map.Mask;
 
 public class VisualDebugger {
 
-	public static final boolean ENABLED = false;// when enabled, mask concurrency is disabled
+	/** When enabled, mask concurrency is disabled.
+	 */
+	public static final boolean ENABLED = false;
 	
-	// If false, color representation of float masks is scaled to include negative ranges.
-	// If true, all negative values are colored as checkerboard, leaving more color space
-	// for positive numbers.
+	/** If false, color representation of float masks is scaled to include negative ranges.
+	 * If true, all negative values are colored as checkerboard, leaving more color space
+	 * for positive numbers.
+	 */
 	public static boolean ignoreNegativeRange = false;
 	
 	private static boolean isDrawAllMasks = false;
--- a/src/java/generator/VisualDebugger.java
+++ b/src/java/generator/VisualDebugger.java
@@ -1,101 +1,142 @@
 package generator;
 
 import java.awt.Color;
-import java.awt.Dimension;
-import java.awt.Graphics;
-import java.awt.Graphics2D;
-import java.awt.GraphicsConfiguration;
-import java.awt.Insets;
-import java.awt.event.WindowEvent;
 import java.awt.image.BufferedImage;
 import java.awt.image.ColorModel;
-
-import javax.swing.JFrame;
-import javax.swing.JPanel;
-import javax.swing.WindowConstants;
+import java.util.HashSet;
+import java.util.Set;
 
 import map.BinaryMask;
 import map.FloatMask;
+import map.Mask;
 
 public class VisualDebugger {
 
-	public static boolean ENABLED = false;// disables mask concurrency
-	public static int perPixelSize = 3;// scale factor for debug image
+	public static boolean ENABLED = false;// when enabled, mask concurrency is disabled
 	
-	private static boolean isRecording = false;
+	// If true, color representation of float masks is scaled to include negative ranges.
+	// If false, all negative values are colored as checkerboard, leaving more color space
+	// for positive numbers.
+	public static boolean ignoreNegativeRange = false;
+	
+	private static boolean isRecordingAllMasks = false;
 	private static BufferedImage currentImage = new BufferedImage(1, 1, BufferedImage.TYPE_INT_RGB);
-	private static GraphicsConfiguration gc;
-	private static JFrame frame;
+	private static Set<Integer> whitelistMasks = null;
 	
-	// PUBLIC
-	
-	static void startRecording() {
-		isRecording = true;
-		if (frame == null) {
-			frame = new JFrame(gc);
-			JPanel panel = new JPanel() {
-				@Override
-		        protected void paintComponent(Graphics g) {				
-		            super.paintComponent(g);
-		            Graphics2D g2d = (Graphics2D) g.create();
-	                int x = (getWidth() - currentImage.getWidth()) / 2;
-	                int y = (getHeight() - currentImage.getHeight()) / 2;
-	                g2d.drawImage(currentImage, x, y, this);
-	                g2d.dispose();
-		        }
-				@Override
-				public Dimension getPreferredSize() {
-					return new Dimension(currentImage.getWidth(), currentImage.getHeight());
-				}
-			};
-			frame.add(panel);
-			frame.setVisible(true);
-			frame.setDefaultCloseOperation(WindowConstants.EXIT_ON_CLOSE);
-			setFrameSize();
+	public static void whitelistMask(Mask binaryOrFloatMask) {
+		if (whitelistMasks == null) {
+			whitelistMasks = new HashSet<>();
 		}
+		whitelistMasks.add(binaryOrFloatMask.hashCode());
 	}
 	
-	static void stopRecording() {
-		isRecording = false;
+	public static void start() {
+		VisualDebuggerGui.createGui(new VisualDebuggerGui.ImagePanel(() -> currentImage, 10));
+	}
+	
+	public static void startRecordAll() {
+		isRecordingAllMasks = true;
+	}
+	
+	public static void stopRecordAll() {
+		isRecordingAllMasks = false;
 	}
 	
 	public static void visualizeMask(BinaryMask mask) {
-		if (!ENABLED || !isRecording) return;
+		if (!shouldRecord(mask)) {
+			return;
+		}
 		visualize((x, y) -> {
 			return mask.get(x, y) ? Color.BLACK.getRGB() : Color.WHITE.getRGB();
-		}, mask.getSize());
+		}, mask.getSize(), "" + mask.hashCode());
 	}
 	
 	public static void visualizeMask(FloatMask mask) {
-		if (!ENABLED || !isRecording) return;
-		// TODO untested
+		if (!shouldRecord(mask)) {
+			return;
+		}
 		visualize((x, y) -> {
 			float value = mask.get(x, y);
 			// if |value| <  1, scale white to black
 			// if |value| 1-10, scale green to red
-			// if |value| > 10, just output blue
-			if (value >= -1 && value <= 1) {
-				float normalized = ((value + 1) / 2);// map to 0 to 1
-				int singleColor = ((int) (normalized * 255)) & 0xFF;
-				return 0xFF_00_00_00
-						& (singleColor << 16)
-						& (singleColor << 8)
-						& (singleColor);
+			// if |value| 10-50, dark blue to blue
+			// if |value| > 100, just output blue
+			
+			if (ignoreNegativeRange && value < 0) {
+				// checkerboard
+				return (x + y) % 2 == 0 ?
+						0xFF_66_66_66 : 0xDD_DD_DD_DD ;
 			}
-			else if (value >= -10 && value <= 10) {
-				float normalized = ((value + 10) / 20);// map to 0 to 1
-				int green = ((int) (255 - (normalized * 255))) & 0xFF;
-				int red = ((int) (normalized * 255)) & 0xFF;
-				return 0xFF_00_00_00
-						& (red << 16)
-						& (green << 8);
+			// For each range we interpolate from color white to the color given in the corresponding mask.
+			// Ranges must be ordered smallest to biggest. Anything bigger than biggest range becomes black.
+			// Anything smaller than smallest becomes checkerboard.
+			int[] colors = new int[] {
+					0xFF_FF_00_00,
+					0xFF_00_FF_00,
+					0xFF_00_00_FF,
+					0xFF_FF_00_FF,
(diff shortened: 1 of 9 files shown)


Katharsas:
@Sheikah45 heyho

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+import map.BinaryMask;
+import map.FloatMask;
+
+public class VisualDebugger {
+
+	public static boolean ENABLED = false;// disables mask concurrency
+	public static int perPixelSize = 3;// scale factor for debug image
+	
+	private static boolean isRecording = false;
+	private static BufferedImage currentImage = new BufferedImage(1, 1, BufferedImage.TYPE_INT_RGB);
```
Geosearchef: I'd choose ARGB.

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+	private static GraphicsConfiguration gc;
+	private static JFrame frame;
+	
+	// PUBLIC
```
Geosearchef: ?

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+	
+	static void startRecording() {
+		isRecording = true;
```
Geosearchef: indent

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+		if (frame == null) {
```
Geosearchef:
```
if(frame != null) {
    return;
}```
Saves the entire indent

File: src/java/generator/VisualDebugger.java [hunk 5]
```diff
@@ -0,0 +1,136 @@
...
+			frame = new JFrame(gc);
+			JPanel panel = new JPanel() {
+				@Override
+		        protected void paintComponent(Graphics g) {
```
Geosearchef: indent

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+	static void startRecording() {
+		isRecording = true;
+		if (frame == null) {
+			frame = new JFrame(gc);
+			JPanel panel = new JPanel() {
+				@Override
+		        protected void paintComponent(Graphics g) {				
+		            super.paintComponent(g);
+		            Graphics2D g2d = (Graphics2D) g.create();
+	                int x = (getWidth() - currentImage.getWidth()) / 2;
```
Geosearchef: indent

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+			setFrameSize();
+		}
+	}
+	
+	static void stopRecording() {
+		isRecording = false;
+	}
+	
+	public static void visualizeMask(BinaryMask mask) {
+		if (!ENABLED || !isRecording) return;
```
Geosearchef: add brackets and newline

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+		visualize((x, y) -> {
+			return mask.get(x, y) ? Color.BLACK.getRGB() : Color.WHITE.getRGB();
+		}, mask.getSize());
+	}
+	
+	public static void visualizeMask(FloatMask mask) {
+		if (!ENABLED || !isRecording) return;
+		// TODO untested
```
Geosearchef: test :)

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+	private interface ImageSource {
+		/**
+		 * @return Color (A)RGB, see {@link ColorModel#getRGBdefault()}, alpha will be ignored.
+		 */
+		public int get(int x, int y);
+	}
+	
+	private static void visualize(ImageSource imageSource, int size) {
+		int imageSize = size * perPixelSize;
+		currentImage = new BufferedImage(imageSize, imageSize, BufferedImage.TYPE_INT_RGB);
```
Geosearchef: ARGB?

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,136 @@
...
+		// iterate source pixels
+		for (int y = 0; y < size; y++) {
```
Geosearchef, line 151: try replacing this loop block with some stdlib functionality. You're basically resizing and painting an image.

src/java/map/BinaryMask.java
```diff
@@ -65,6 +67,7 @@ public BinaryMask randomize(float density) {
 			}
 		}
 		applySymmetry();
+		VisualDebugger.visualizeMask(this);
```
Geosearchef, line 70: what if instead of duplicating this a lot of times we do it in the concurrent mask that is calling this?

src/java/map/ConcurrentBinaryMask.java
```diff
@@ -33,90 +35,150 @@ public ConcurrentBinaryMask(ConcurrentBinaryMask mask, long seed, String name) {
 	}
 
 	public ConcurrentBinaryMask randomize(float density) {
+		if (VisualDebugger.ENABLED) {
+			binaryMask.randomize(density);
+			return this;
+		}
```
Geosearchef: What if instead of duplicating this a lot of times this call get's moved to the Pipeline.add() function?

src/java/map/ConcurrentFloatMask.java
```diff
@@ -33,43 +35,71 @@ public ConcurrentFloatMask(ConcurrentFloatMask mask, long seed, String name) {
...
 				this.floatMask.add(((ConcurrentFloatMask)res.get(1)).getFloatMask())
 		);
 	}
 
 	public ConcurrentFloatMask max(ConcurrentBinaryMask other) {
+		if (VisualDebugger.ENABLED) {
+			throw new UnsupportedOperationException();
+		}
 		return Pipeline.add(this, Arrays.asList(this, other), res ->
+				// Probably a bug, res should be BinaryMask
```
Geosearchef: probably correct

Geosearchef (changes requested):
Really nice!
This approach won't work though in case the visualized part depends on something that should be run in parallel beforehand, as that would only have been executed after `Pipeline.start()` was called:
I also think this will break any concurrent computations coming later. 

What do you think about moving some of this code to `Pipeline.java`? This way duplicate code could be reduced a lot. Maybe we should also consider disabling concurrency completely in there when visualizations are enabled.

Another suggestion which I think would be the cleanest is to add a visualize operation to the concurrent masks. (or add another stage to the completableFuture in `Pipeline`) This way a visualize call is part of the pipeline as any other operation on a mask. Inside of that you can then run the actual visualization and pause till the users proceeds to the next step, therefore having all operations that depend on the visualized one wait.

src/java/generator/VisualDebugger.java
Katharsas: Well RGB is basically the same as ARGB except that alpha channel is ignored, which is what we want i think.

src/java/generator/VisualDebugger.java
Katharsas: I don't think there is an STD function that takes an arbitrary pixel source lamda and constructs an image from that.
And if there was, it would double the amount of memory copies but that would be ok for debugging i guess.

src/java/map/BinaryMask.java
Katharsas: Then it wouldn't work with non-concurrent masks right?

Geosearchef: Hmm, right.

Geosearchef: That would still work for the "add visualization as a method to masks" approach though.

Katharsas:
@Geosearchef
The idea was that it should be trivial to correlate which function call corresponds to which map visualization. The current implementation  basically converts concurrent masks back to non-concurrent masks while the debugger is enabled.
This way, when you breakpoint and step to a certain function call like `plateaus.combine(mountains);`, then you can see what this call does rigth there. So this is why i decided to basically remove concurrency while debugging.

Of course, you could also breakpoint to the section where the pipeline is executed, but then you have to step through all steps until you actually find the one that you are interested in and i don't like that idea for debugging.

So im in favour of not involving the piping code for simplicity and ease of debugging.

To remove the duplication, maybe we could use annotations or something to automatically instrument all relevant functions calls instead of manually intrumenting them, but that seems toi much work just for deduplicating debugging code.

src/java/generator/VisualDebugger.java
Katharsas: removed

src/java/generator/VisualDebugger.java
Katharsas: ok

src/java/map/ConcurrentBinaryMask.java
Katharsas: see general comment on the PR

src/java/map/ConcurrentFloatMask.java
Katharsas: can you just push the fix on master or smthn, might not want to mix it with this PR

Katharsas:
Implemented float mask coloring properly with customizable number of color layers and ranges:
Annotation 2020-06-16 015351

In realtime: https://www.youtube.com/watch?v=ouGVq9mFFKc

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,177 @@
...
+import map.FloatMask;
+import map.Mask;
+
+public class VisualDebugger {
+
+	public static final boolean ENABLED = false;// when enabled, mask concurrency is disabled
+	
+	// If false, color representation of float masks is scaled to include negative ranges.
+	// If true, all negative values are colored as checkerboard, leaving more color space
+	// for positive numbers.
```
1-alex98: Java docs

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import map.BinaryMask;
+import map.FloatMask;
+import map.Mask;
+
+public class VisualDebugger {
+
+	public static final boolean ENABLED = false;// when enabled, mask concurrency is disabled
```
1-alex98: java docs

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,177 @@
...
+	
+	// If false, color representation of float masks is scaled to include negative ranges.
+	// If true, all negative values are colored as checkerboard, leaving more color space
+	// for positive numbers.
+	public static boolean ignoreNegativeRange = false;
+	
+	private static boolean isDrawAllMasks = false;
+	private static Set<Integer> drawMasksWhitelist = null;// whitelisted masks are always drawn
```
1-alex98: java docs

src/java/generator/VisualDebugger.java
```diff
@@ -0,0 +1,177 @@
...
+import java.awt.image.BufferedImage;
+import java.awt.image.ColorModel;
+import java.util.HashSet;
+import java.util.Set;
+
+import map.BinaryMask;
+import map.FloatMask;
+import map.Mask;
+
+public class VisualDebugger {
```
1-alex98, line 13: Singelton

src/java/generator/VisualDebuggerGui.java
```diff
@@ -0,0 +1,167 @@
...
+		@Override
+        protected void paintComponent(Graphics g) {
+            super.paintComponent(g);
+            Graphics2D g2d = (Graphics2D) g.create();
+            BufferedImage currentImage = image;
+            int x = (getWidth() - currentImage.getWidth()) / 2;
+            int y = (getHeight() - currentImage.getHeight()) / 2;
+            g2d.drawImage(currentImage, x, y, this);
+            g2d.dispose();
+        }
```
1-alex98: format

src/java/generator/VisualDebuggerGui.java
```diff
@@ -0,0 +1,167 @@
...
+		public boolean equals(Object other) {
+			if (other instanceof MaskListItem) {
+				return this.maskName.equals(((MaskListItem)other).maskName);
+			}
+			return false;
+		}
+		@Override
+		public String toString() {
+			return "  " + maskName + "  ";
+		}
```
1-alex98: Lombok

1-alex98 (changes requested):

1-alex98 (approved):

src/java/generator/VisualDebugger.java
Katharsas: no

src/java/generator/VisualDebugger.java
Katharsas: done

src/java/generator/VisualDebugger.java
Katharsas: done

src/java/map/BinaryMask.java
Katharsas: What do you mean?

src/java/map/ConcurrentFloatMask.java
Katharsas: ok i removed the method because i makes no sense

src/java/generator/VisualDebuggerGui.java
```diff
@@ -0,0 +1,156 @@
...
+			this.zoomFactor = zoomFactor;
+		}
+		
+		public int getZoomFactor() {
+			return zoomFactor;
+		}
+		
+		@Override
+		protected void paintComponent(Graphics g) {
+            super.paintComponent(g);
```
Katharsas: whitespace

Katharsas: fixed

1-alex98:
Should we merge this?

Katharsas:
I don't plan to do additional work here, so from my point of view yes.

1-alex98:
merged on accident but since we wanna merge it anyways

PR #32, opened 2020-06-15
Changed Rock Reclaim Generation

Sheikah45:
Small rocks are generated around plateaus and Mountains and removed large rock circle in center.

Issue #28, opened 2020-05-26
Random reclaim generation


Katharsas:
Seems to currently always generate a lot of big rocks towards center of map.

Issue #8, opened 2018-05-06
Generate neutral structures


n3roxis:
Civilian structures and abandoned military bases give maps character.

PR #33, opened 2020-06-16
Added in Wrecks and Civilian Structures

### LINKED ISSUE #41: Added in Wrecks and Civilian Structures

Sheikah45:
Added in Representations for Civilian Units and Wrecks under the Unit Class. Add in generators for both Civilian Units and Wrecks. Wrecks currently include t1,t2, and t3 land, t2 navy, and navy factories. Added in mask generation for wreck placement on map. Civilian Mask and bp files not included.
Civilians will just need more structure to their placement but the framework is there to do it.

Also fixed rotation mirroring for props.


Sheikah45:
Could potentially be merged with #32 but left it as its own due to new class creation.

Sheikah45:
Addresses #28 and #8

src/java/generator/MapGenerator.java
```diff
@@ -170,6 +172,30 @@ public SCMap generate(long seed) throws ExecutionException, InterruptedException
...
+
+		wreckGenerator.generateWrecks(t1LandWreckMask, wreckGenerator.T1_Land, 2f);
+		wreckGenerator.generateWrecks(t2LandWreckMask, wreckGenerator.T2_Land, 15f);
+		wreckGenerator.generateWrecks(t3LandWreckMask, wreckGenerator.T3_Land, 30f);
+		wreckGenerator.generateWrecks(t2NavyWreckMask, wreckGenerator.T2_Navy, 60f);
+		wreckGenerator.generateWrecks(navyFactoryWreckMask, wreckGenerator.Navy_Factory, 120f);
+
+		BinaryMask allWreckMask = new BinaryMask(513, random.nextLong());
+		allWreckMask.combine(t1LandWreckMask).combine(t2LandWreckMask).combine(t3LandWreckMask).combine(t2NavyWreckMask).inflate(2);
+		noProps.combine(allWreckMask);
```
1-alex98: Extracting that into a method would be really cool :D

src/java/generator/MapGenerator.java
```diff
@@ -32,8 +32,10 @@
 	public static final String VERSION = "0.1.6";
 
 	//read from cli args
+//	private static Random randomseed = new Random();
 	private static String FOLDER_PATH = ".";
 	private static String MAP_NAME = "debugMap";
+//	private static long SEED = randomseed.nextLong();
```
1-alex98: Let's not leave dead code around... just delete it

src/java/generator/UnitGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+import map.SCMap;
+import util.Vector2f;
+import util.Vector3f;
+
+import java.util.Random;
+
+public strictfp class UnitGenerator {
+    public final String[] STRUCTURES = {
+
+    };
```
1-alex98: is that used?

src/java/generator/WreckGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+            "UAL0303",
+            "XSL0303"
+    };
+    public final String[] T2_Navy = {
+            "UES0201",
+            "URS0201",
+            "XSS0201",
+            "UAS0201"
+    };
+    public final String[] Navy_Factory = {
```
1-alex98: constants are normally static

src/java/generator/UnitGenerator.java
Sheikah45: This is not used in this iteration however I left it in since this will be the place that civilian structures are added so they can be placed easily.

src/java/generator/MapGenerator.java
Sheikah45: I can put this into a method but had left it out since all the other prop and mask generation is done directly in the generate function and this generation does rely on a few few of those masks. I could also clean up the ordering as well.

src/java/generator/MapGenerator.java
Sheikah45: Will do

src/java/generator/WreckGenerator.java
Sheikah45: Done

build.gradle
```diff
@@ -29,6 +29,7 @@ repositories {
 
 dependencies {
     compileOnly 'org.projectlombok:lombok:1.18.8'
+    annotationProcessor 'org.projectlombok:lombok:1.18.8'
```
Geosearchef: This will conflict with your other PR

src/java/map/Unit.java
```diff
@@ -0,0 +1,27 @@
...
+        return type;
+    }
+
+    public Vector3f getPosition() {
+        return position;
+    }
+
+    public float getRotation() {
+        return rotation;
+    }
```
Geosearchef: use lombok's `@Getter` (in front of class or each attribute)

1-alex98 (approved):

PR #34, opened 2020-06-16
Add Licence

1-alex98:
Took my name as I was not sure if u wanna be mentioned with your real name. 

Please confirm you wanna add MIT Liscene to this project.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Alexander von Trostorff
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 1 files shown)


n3roxis:
I think using my real name is appropriate. You can do so.

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Alexander von Trostorff
```
n3roxis: Copyright (c) 2020 Peter Ziegler

n3roxis (changes requested):

1-alex98:
@Geosearchef @Rackover a small message that u are ok with that would be awesome :D

Rackover:
We can have multiple names on the license, so I do not mind whether or not only Neroxis' name is credited or if we put our three/four names on the License. I think it is up to Neroxis to decide.

Also, you do not need my permission to do something for FAF 😛
Please stop pinging me, my mail box is already getting a lot of notifications from FAF github that I must disable every time independantly. 

If you need my name for credits purposes, use Roxanne De Louvegarde <rackover@racknet.noip.me>

1-alex98:
Sure u need to agree to a license

n3roxis (approved):

PR #37, opened 2020-06-19
Add in Argument Parsing with options seed

### LINKED ISSUE #29: Make spawn number configurable

### LINKED ISSUE #35: Long term argument handling / client integration
Summary of what we talked about in Discord:

The map name should mostly consist of 3 concatinated parts:
- MapGen version fixed format.
- Options (key value map) encoded as string to be shorter:
  - Seed for PRNG
  - Number of spots
  - etc. (ideas: amount of reclaim or map size or amount of water/hills), should be easy to extend

### Creating a generator game:
When creating a game, the user will have to input a bunch of options for map generation. This requires UI implementation in client. This means that client always hosts games for a specific version of the map gen that its UI was written for regardless of if newer version of mapgen exists.

The UI should eventually provide the possibility to accept a map name to predefine seed and options so a previously played map can be played again, with same or different options. Client should either reject map names from too new mapgen versions or disable the option-UI in that case.

When player wants to play the map, the generator re-encodes seed and options and passes map name to mapgen so it can generate map.

### Joining a generator game:
When joining, the client should just pass the entire name of the map, including the encoded options to the mapgen. The encocded string should essentially be a blackbox to the client in this case.

### Problems with getting this to work:

Right now the client verifies format of map name even when just joining a game or playing replay, which means the map name format cannot currently change without client change. The mapgen on the other hand, needs to be able to decode the options string by itself.

### Possible solution:
The goal is to have a proper way to make make similarly big changes in the future:   
The map gen specifies a major and a minor version number. When the minor number changes, the rules above for creating and joining apply. When the major number changes, the client will not just ignore the newer mapgen when hosting, it will also refuse to replay and join generator games and tell the player to update his client.  
This should be implemented first. Then we can implement the changes above for mapgen 2.0, and any map names that are > 2.0 will cause the client to refuse to even look at the map name and require an update. Then we release a new client that implements handling of >= 2.0 map names outlined above.

### Disadvantages:
Mapgen needs to implement option decoding, client needs to implement decoding and encoding for the version for which it has UI support, so we have duplicated decoding code.

Feedback welcome.

Sheikah45:
This looks to attempt to solve #35 with including a random seed and an options seed in the map name. Both seeds are encoded as base64 strings to try to keep the length of the map names reasonable. 

Currently each argument is a single byte which is then translated into an appropriate value in the parse options method. The original command line argument parser is adapted and left intact to minimize the changes necessary to the client to send over arguments as well as to try and increase compatibility between map generator versions. Default values are defined in case the options string is not passed and an encoding function for the map name is included for generation of the map name solely inside of the map generator if needed. 

Additionally other argument parsing methods have been added so that developers do not have to have an encoded string for testing however a valid map name is generated and used to save the map with all argument parsing options. For all argument parsing the map name will override any other options that are set.

The base64 encoding is changed slightly when saving the map by replacing the '\' character with a '-' character. When parsing the map the '-' character is replaced with a '\' character to ensure a base64 compliant string in the case that the slightly modified version is given to the generator.


Sheikah45:
This also deals with FAForever/downlords-faf-client#1781

Sheikah45:
As we discussed this should also move the version up to 1.0.0

1-alex98 (approved):

PR #38, opened 2020-06-21
Remove and ignore Eclipe specific files

Katharsas:
These are not needed.
Import project through gradle buildship plugin instead of using "add existing project" when using Eclipse.


Sheikah45 (approved):

1-alex98 (approved):

1-alex98:
@n3roxis we still can not merge

PR #67, opened 2020-07-07
Parameterize Map Size

### LINKED ISSUE #45: Generate maps of varying size
Currently the map generator only supports generation of 10x10 maps. Support should be extended for generating viable maps of any size and added as a potential option at generation time.

Sheikah45:
Allows for maps of different sizes to be generated as an option.

Fixes #45

### DIFF
--- a/src/java/export/SCMapExporter.java
+++ b/src/java/export/SCMapExporter.java
@@ -2,6 +2,7 @@ package export;
 
 import map.SCMap;
 import map.TerrainMaterials;
+import util.DDSHeader;
 import util.Vector2f;
 import util.Vector3f;
 import util.Vector4f;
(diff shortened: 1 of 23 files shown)


src/java/export/SCMapExporter.java
```diff
@@ -186,7 +219,7 @@ public static void exportSCMAP(Path folderPath, String mapname, SCMap map) throw
 			ImageIO.write(renderedImage, fileFormat, previewFile);
 		}
 		catch (IOException e) {
-			System.out.printf("Could not write the preview image\n"+e.toString());
+			System.out.print("Could not write the preview image\n" + e.toString());
```
1-alex98: Strange way of logging exceptions

1-alex98 (approved):

PR #209, opened 2020-10-20
Added numerous functions, improved waterBeach and related, and added options

Penguin5:
Added removeValuesOutsideOf, removeValuesInsideOf, removeValuesOutsideOfRange, removeValuesInRange, replaceValuesInRangeWith, copyWithinRange, copyOutsideRange, smoothWithinSpecifiedDistanceOfEdgesOf, reduceValuesOnIntersectingSmoothingZones, convertToBinaryMask, getDistanceFieldForRange, removeAreasOutsideOfSpecifiedIntensityAndSize, removeAreasOfSpecifiedIntensityAndSize, and removeAreasOfSpecifiedLocalMaximums FloatMask functions
Added removeAreasBiggerThan, removeAreasOutsideOfSpecifiedSize, removeAreasOfSpecifiedSize, getAreasWithinSpecifiedDistanceOfEdges, and setRectangularAreaWithoutSymmetry BinaryMask functions
Fixed waterBeachTexture going onto ridges
Fixed waterBeachTexture going inland so much
Made waterBeachTexture less present on narrow lands
Modified populator texturing for small bodies of water
Added an option to populator to change texture layer for small bodies of water
Added populator options for only populating textures in certain areas
Renamed filterShapes function to removeAreasSmallerThan
Added smallWaterSizeLimit variable
Misc minor formatting changes

### DIFF
--- a/src/java/map/BinaryMask.java
+++ b/src/java/map/BinaryMask.java
@@ -762,7 +762,7 @@ public strictfp class BinaryMask extends Mask {
         return this;
     }
 
-    public BinaryMask filterShapes(int minArea) {
+    public BinaryMask removeAreasSmallerThan(int minArea) {
         LinkedHashSet<Vector2f> locHash = new LinkedHashSet<>();
         FloatMask distanceField = getDistanceField();
         for (int x = 0; x < getSize(); x++) {
@@ -803,6 +803,13 @@ public strictfp class BinaryMask extends Mask {
             locHash.removeAll(coordinates);
             locList = new LinkedList<>(locHash);
         }
+        applySymmetry(symmetrySettings.getSpawnSymmetry());
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public BinaryMask removeAreasBiggerThan(int maxArea) {
+        minus(copy().removeAreasSmallerThan(maxArea));
         applySymmetry(symmetrySettings.getSpawnSymmetry());
         VisualDebugger.visualizeMask(this);
         return this;
--- a/src/java/map/BinaryMask.java
+++ b/src/java/map/BinaryMask.java
@@ -852,7 +852,6 @@ public strictfp class BinaryMask extends Mask {
 
     public BinaryMask getAreasWithinSpecifiedDistanceOfEdges(int distance) {
         BinaryMask newMask = copy().inflate(distance).minus(copy().deflate(distance));
-        VisualDebugger.visualizeMask(this);
         return newMask;
     }
 
--- a/src/java/map/BinaryMask.java
+++ b/src/java/map/BinaryMask.java
@@ -134,6 +134,39 @@ public strictfp class BinaryMask extends Mask {
                 || (y < getSize() - 1 && get(x, y + 1) != value));
     }
 
+    public BinaryMask setRectangularAreaWithoutSymmetry(int x1, int x2, int z1, int z2, boolean trueOrFalse) {
+        int smallX;
+        int bigX;
+        int smallY;
+        int bigY;
+        if (x1 > x2) {
+            smallX = x2;
+            bigX = x1;
+        } else {
+            smallX = x1;
+            bigX = x2;
+        }
+        if (z1 > z2) {
+            smallY = z2;
+            bigY = z1;
+        } else {
+            smallY = z1;
+            bigY = z2;
+        }
+        for (int x = 0; x < getSize(); x++) {
+            for (int y = 0; y < getSize(); y++) {
+                if (x > smallX && x < bigX && y > smallY && y < bigY)
+                    if (trueOrFalse) {
+                    set(x, y, false);
+                    } else {
+                    set(x, y, true);
+                    }
+            }
+        }
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
     public BinaryMask copy() {
         if (random != null) {
             return new BinaryMask(this, random.nextLong());
--- a/src/java/map/BinaryMask.java
+++ b/src/java/map/BinaryMask.java
@@ -134,7 +134,7 @@ public strictfp class BinaryMask extends Mask {
                 || (y < getSize() - 1 && get(x, y + 1) != value));
     }
 
-    public BinaryMask setRectangularAreaWithoutSymmetry(int x1, int x2, int z1, int z2, boolean trueOrFalse) {
+    public BinaryMask setRectangularAreaFromPoints(int x1, int x2, int z1, int z2, boolean trueOrFalse) {
         int smallX;
         int bigX;
         int smallY;
@@ -153,18 +153,7 @@ public strictfp class BinaryMask extends Mask {
             smallY = z1;
             bigY = z2;
         }
-        for (int x = 0; x < getSize(); x++) {
-            for (int y = 0; y < getSize(); y++) {
-                if (x > smallX && x < bigX && y > smallY && y < bigY)
-                    if (trueOrFalse) {
-                    set(x, y, false);
-                    } else {
-                    set(x, y, true);
-                    }
-            }
-        }
-        VisualDebugger.visualizeMask(this);
-        return this;
+        return fillParallelogram(smallX, smallY, bigX - smallX, bigY - smallY, 0, 0, trueOrFalse);
     }
 
     public BinaryMask copy() {
--- a/src/java/map/BinaryMask.java
+++ b/src/java/map/BinaryMask.java
@@ -815,6 +815,25 @@ public strictfp class BinaryMask extends Mask {
         return this;
     }
 
+    public BinaryMask removeAreasOutsideOfSpecifiedSize(int minSize, int maxSize) {
+        removeAreasSmallerThan(minSize);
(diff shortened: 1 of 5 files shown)


src/java/map/FloatMask.java
```diff
@@ -422,6 +422,118 @@ public FloatMask setSize(int size) {
...
+    public FloatMask replaceValuesInRangeWith(BinaryMask range, FloatMask replacement) {
+        removeValuesInsideOf(range).add(replacement.copy().removeValuesOutsideOf(range));
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public FloatMask copyWithinRange(BinaryMask other) {
+        FloatMask newMask = copy().removeValuesOutsideOf(other);
+        VisualDebugger.visualizeMask(this);
+        return newMask;
```
Sheikah45, line 473: Is this supposed to return newMask or is it supposed to modify this mask?
Also this would break the visualizer logic if we are expecting the same mask back.
Aslo the call to visualize mask doesnt matter currently as this is never changed

src/java/map/FloatMask.java
```diff
@@ -422,6 +422,118 @@ public FloatMask setSize(int size) {
...
+    }
+
+    public FloatMask copyOutsideRange(BinaryMask other) {
+        FloatMask newMask = copy().removeValuesInsideOf(other);
+        VisualDebugger.visualizeMask(this);
+        return newMask;
```
Sheikah45, line 478: same as copywithinRange

src/java/map/FloatMask.java
Penguin5: return new mask
visualize can be removed if there's no good way to do that

src/java/map/FloatMask.java
Penguin5: same response

src/java/map/BinaryMask.java
```diff
@@ -134,6 +134,39 @@ public boolean isEdge(int x, int y) {
                 || (y < getSize() - 1 && get(x, y + 1) != value));
     }
 
+    public BinaryMask setRectangularAreaWithoutSymmetry(int x1, int x2, int z1, int z2, boolean trueOrFalse) {
```
Sheikah45: this function already exists. See fillRect and for more arbitrary see FillParallelogram

Sheikah45 (changes requested):

src/java/map/BinaryMask.java
```diff
@@ -808,6 +830,32 @@ public BinaryMask filterShapes(int minArea) {
...
+
+    public BinaryMask removeAreasOfSpecifiedSize(int minSize, int maxSize) {
+        minus(this.copy().removeAreasOutsideOfSpecifiedSize(minSize, maxSize));
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public BinaryMask getAreasWithinSpecifiedDistanceOfEdges(int distance) {
+        BinaryMask newMask = copy().inflate(distance).minus(copy().deflate(distance));
+        VisualDebugger.visualizeMask(this);
```
Sheikah45: Visualize call her does nothing

src/java/map/FloatMask.java
```diff
@@ -422,6 +422,116 @@ public FloatMask setSize(int size) {
...
+    public FloatMask copyOutsideRange(BinaryMask other) {
+        FloatMask newMask = copy().removeValuesInsideOf(other);
+        return newMask;
+    }
+
+    public FloatMask smoothWithinSpecifiedDistanceOfEdgesOf(BinaryMask other, int distance) {
+        for (int x = 0; x < distance; x = x + 2) {
+            replaceValuesInRangeWith(other.getAreasWithinSpecifiedDistanceOfEdges(x + 1), copy().smooth(1));
+        }
+        VisualDebugger.visualizeMask(this);
```
Sheikah45, line 485: Does this need to be in a for loop?
At the surface the for loop appears unnecessary as the values are just replaced seems like wasted effort

src/java/map/FloatMask.java
```diff
@@ -422,6 +422,116 @@ public FloatMask setSize(int size) {
...
+
+    public FloatMask reduceValuesOnIntersectingSmoothingZones(BinaryMask avoidMakingZonesHere) {
+        FloatMask newMask = copy().smooth(34).subtract(copy()).subtract(avoidMakingZonesHere, 1f);
+        BinaryMask Zones = newMask.copy().removeValuesInRange(0f, 0.5f).smooth(2).convertToBinaryMask(0.5f, 1f).inflate(34);
+        BinaryMask newMaskBase = convertToBinaryMask(1f, 1f).deflate(3).minus(Zones.copy().invert());
+        newMask.init(newMaskBase, 0, 1).smooth(4).clampMax(0.35f).add(newMaskBase, 1f).smooth(2).clampMax(0.65f).add(newMaskBase, 1f).smooth(1).add(newMaskBase, 1f).clampMax(1f);
+        replaceValuesInRangeWith(Zones, newMask).smoothWithinSpecifiedDistanceOfEdgesOf(Zones, 30);
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
```
Sheikah45, line 497: This function assumes the float is only 0 to 1 this limits its utility

src/java/map/FloatMask.java
```diff
@@ -422,6 +422,116 @@ public FloatMask setSize(int size) {
...
+    public FloatMask smoothWithinSpecifiedDistanceOfEdgesOf(BinaryMask other, int distance) {
+        for (int x = 0; x < distance; x = x + 2) {
+            replaceValuesInRangeWith(other.getAreasWithinSpecifiedDistanceOfEdges(x + 1), copy().smooth(1));
+        }
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public FloatMask reduceValuesOnIntersectingSmoothingZones(BinaryMask avoidMakingZonesHere) {
+        FloatMask newMask = copy().smooth(34).subtract(copy()).subtract(avoidMakingZonesHere, 1f);
```
Sheikah45: can this be named not newMask as it is not returned

src/java/map/FloatMask.java
```diff
@@ -422,6 +422,116 @@ public FloatMask setSize(int size) {
...
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public FloatMask removeAreasOutsideOfSpecifiedIntensityAndSize(int minSize, int maxSize, float minIntensity, float maxIntensity) {
+        FloatMask tempMask2 = copy().init(this.copy().convertToBinaryMask(minIntensity, maxIntensity).removeAreasOutsideOfSpecifiedSize(minSize, maxSize).invert(), 0f, 1f);
+        this.subtract(tempMask2).clampMin(0f);
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
```
Sheikah45, line 516: same limited utility due to range assumptions

src/java/map/FloatMask.java
```diff
@@ -422,6 +422,116 @@ public FloatMask setSize(int size) {
...
+    public FloatMask removeAreasOfSpecifiedIntensityAndSize(int minSize, int maxSize, float minIntensity, float maxIntensity) {
+        subtract(this.copy().removeAreasOutsideOfSpecifiedIntensityAndSize(minSize, maxSize, minIntensity, maxIntensity));
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public FloatMask removeAreasOfSpecifiedSizeWithLocalMaximums(int minSize, int maxSize, int levelOfPrecision) {
+        for (int x = 0; x < levelOfPrecision; x++) {
+            removeAreasOfSpecifiedIntensityAndSize(minSize, maxSize, (1f -(float) x / (float) levelOfPrecision), 1f);
+        }
```
Sheikah45, line 527: why the for loop?

Sheikah45 (changes requested):
Also this needs to be rebased

src/java/map/FloatMask.java
Penguin5: so you can specify level of precision
it cleans up texturing spreading (such as behind ridges) at basically equidistant arbitrary levels
more precision = more arbitrary levels = cleaner/fuller/more precise removal

src/java/map/BinaryMask.java
Penguin5: will remove

src/java/map/FloatMask.java
Penguin5: It's not on the for loop.
It's after the for loop.

Penguin5: Also, in my testing, repeated smoothing of small values often produces predictably different results compared to smoothing once with a small or large value.

Sheikah45: yes I understand but those values will get overridden with the replace function anyway if I understand it correctly

Sheikah45: as every loop iteration you are replacing the values

Sheikah45 (approved):
Looks good

PR #225, opened 2020-10-30
Created new FloatMask/BinaryMask functions related to scaling and or distance fields

Penguin5:
Created ScaleFloatMaskBy, ScaleFloatMaskTo, scaleFloatMaskFromTo, and getRankOfPoint FloatMask functions
Created getDistanceBetweenPoints (3 versions for varying inputs), getNearestPointInAreaTo, getLowestDistanceFromPointToArea, getDistanceFieldFromPoint, getDistanceFieldFromPointInArea, getRankedDistanceFieldFromPoint, getRankedDistanceFieldFromPointInArea, and getDistanceFieldFromArea BinaryMask functions

### DIFF
--- a/src/java/map/BinaryMask.java
+++ b/src/java/map/BinaryMask.java
@@ -3,12 +3,10 @@ package map;
 import generator.VisualDebugger;
 import lombok.Getter;
 import lombok.SneakyThrows;
-import org.luaj.vm2.ast.Str;
 import util.Util;
 import util.Vector2f;
 import util.Vector3f;
 
-import java.awt.*;
 import java.io.BufferedOutputStream;
 import java.io.DataOutputStream;
 import java.io.FileOutputStream;
@@ -18,7 +16,6 @@ import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.*;
-import java.util.List;
 
 @Getter
 public strictfp class BinaryMask extends Mask {
--- a/src/java/map/BinaryMask.java
+++ b/src/java/map/BinaryMask.java
@@ -3,10 +3,12 @@ package map;
 import generator.VisualDebugger;
 import lombok.Getter;
 import lombok.SneakyThrows;
+import org.luaj.vm2.ast.Str;
 import util.Util;
 import util.Vector2f;
 import util.Vector3f;
 
+import java.awt.*;
 import java.io.BufferedOutputStream;
 import java.io.DataOutputStream;
 import java.io.FileOutputStream;
@@ -16,6 +18,7 @@ import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.*;
+import java.util.List;
 
 @Getter
 public strictfp class BinaryMask extends Mask {
@@ -91,6 +94,41 @@ public strictfp class BinaryMask extends Mask {
 
     public void set(int x, int y, boolean value) {
         mask[x][y] = value;
+    }
+
+    public float getDistanceBetweenPoints(int x1, int y1, int x2, int y2) {
+        return (float) StrictMath.sqrt((double) ((x1 - x2) ^ 2 + (y1 - y2) ^ 2));
+    }
+
+    public float getDistanceBetweenPoints(int x1, int y1, Point Point2) {
+        return (float) StrictMath.sqrt((double) ((x1 - Point2.x) ^ 2 + (y1 - Point2.y) ^ 2));
+    }
+
+    public float getDistanceBetweenPoints(Point Point1, Point Point2) {
+        return (float) StrictMath.sqrt((double) ((Point1.x - Point2.x) ^ 2 + (Point2.x - Point2.y) ^ 2));
+    }
+
+    public Point getNearestPointInAreaTo(int x, int y) {
+        Point nearestPoint = new Point (x, y);
+        int rank;
+        FloatMask distanceField = getDistanceFieldFromPointInArea(x, y);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                if (this.get(a, b)) {
+                    rank = distanceField.getRankOfPointInArea(a, b, this);
+                    if(rank == 1) {
+                        nearestPoint.setLocation(a, b);
+                        a = getSize();
+                        b = getSize();
+                    }
+                }
+            }
+        }
+        return nearestPoint;
+    }
+
+    public float getLowestDistanceFromPointToArea(int x, int y) {
+        return getDistanceBetweenPoints(x, y, getNearestPointInAreaTo(x, y));
     }
 
     public boolean isEdge(int x, int y) {
@@ -809,6 +847,62 @@ public strictfp class BinaryMask extends Mask {
     public BinaryMask getAreasWithinSpecifiedDistanceOfEdges(int distance) {
         BinaryMask newMask = copy().inflate(distance).minus(copy().deflate(distance));
         return newMask;
+    }
+
+    public FloatMask getDistanceFieldFromPoint(int x, int y) {
+        FloatMask distanceField = new FloatMask(getSize(), null, symmetrySettings);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                distanceField.set(a, b, getDistanceBetweenPoints(x, y, a, b));
+            }
+        }
+        return distanceField;
+    }
+
+    public FloatMask getDistanceFieldFromPointInArea(int x, int y) {
+        FloatMask distanceField = new FloatMask(getSize(), null, symmetrySettings);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                if (this.get(a, b)) {
+                    distanceField.set(a, b, getDistanceBetweenPoints(x, y, a, b));
+                }
+            }
+        }
+        return distanceField;
+    }
+
+    public FloatMask getRankedDistanceFieldFromPoint(int x, int y) {
+        FloatMask distanceField = getDistanceFieldFromPoint(x, y);
+        FloatMask rankedDistanceField = new FloatMask(getSize(), null, symmetrySettings);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                rankedDistanceField.set(a, b, distanceField.getRankOfPoint(a, b));
+            }
+        }
+        return rankedDistanceField;
+    }
+
+    public FloatMask getRankedDistanceFieldFromPointInArea(int x, int y) {
+        FloatMask distanceField = getDistanceFieldFromPointInArea(x, y);
+        FloatMask rankedDistanceField = new FloatMask(getSize(), null, symmetrySettings);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                if (this.get(a, b)) {
+                    rankedDistanceField.set(a, b, distanceField.getRankOfPointInArea(a, b, this));
+                }
+            }
+        }
+        return rankedDistanceField;
+    }
+
+    public FloatMask getDistanceFieldFromArea() {
+        FloatMask distanceField = new FloatMask(getSize(), null, symmetrySettings);
+        for (int x = 0; x < getSize(); x++) {
+            for (int y = 0; y < getSize(); y++) {
+                distanceField.set(x, y, getLowestDistanceFromPointToArea(x, y));
+            }
+        }
+        return distanceField;
     }
 
     public FloatMask getDistanceField() {
--- a/src/java/map/BinaryMask.java
+++ b/src/java/map/BinaryMask.java
@@ -96,39 +96,27 @@ public strictfp class BinaryMask extends Mask {
         mask[x][y] = value;
     }
 
-    public float getDistanceBetweenPoints(int x1, int y1, int x2, int y2) {
-        return (float) StrictMath.sqrt((double) ((x1 - x2) ^ 2 + (y1 - y2) ^ 2));
-    }
-
-    public float getDistanceBetweenPoints(int x1, int y1, Point Point2) {
-        return (float) StrictMath.sqrt((double) ((x1 - Point2.x) ^ 2 + (y1 - Point2.y) ^ 2));
-    }
-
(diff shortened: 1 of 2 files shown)


src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
         mask[x][y] = value;
     }
 
+    public float getDistanceBetweenPoints(int x1, int y1, int x2, int y2) {
```
Sheikah45: This function shouldn't go here as it is not related to BinaryMask. It should go in utils in some class. One should really just use Vector2f but if you really want a function for just the primitives then put a function in utils

src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
...
+        return (float) StrictMath.sqrt((double) ((x1 - x2) ^ 2 + (y1 - y2) ^ 2));
+    }
+
+    public float getDistanceBetweenPoints(int x1, int y1, Point Point2) {
```
Sheikah45: This shouldnt exist because there is Vector2f.getdistance. Dont want to use point as we already use vector2f

src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
...
+        return (float) StrictMath.sqrt((double) ((x1 - Point2.x) ^ 2 + (y1 - Point2.y) ^ 2));
+    }
+
+    public float getDistanceBetweenPoints(Point Point1, Point Point2) {
```
Sheikah45: same

src/java/map/FloatMask.java
```diff
@@ -572,6 +572,64 @@ public FloatMask removeAreasOfSpecifiedSizeWithLocalMaximums(int minSize, int ma
         return this;
     }
 
+    public FloatMask scaleFloatMaskBy(float scaleMultiplier) {
```
Sheikah45: this is the same as multiply(float)

src/java/map/FloatMask.java
```diff
@@ -572,6 +572,64 @@ public FloatMask removeAreasOfSpecifiedSizeWithLocalMaximums(int minSize, int ma
...
+                multiply(x, y, scaleMultiplier);
+            }
+        }
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public FloatMask scaleFloatMaskTo(float newFloatMax) {
+        float scaleMultiplier = newFloatMax / getMax();
+        for (int x = 0; x < getSize(); x++) {
```
Sheikah45: use multiply(scaleMultiplier) instead of for loops

src/java/map/FloatMask.java
```diff
@@ -572,6 +572,64 @@ public FloatMask removeAreasOfSpecifiedSizeWithLocalMaximums(int minSize, int ma
...
+                multiply(x, y, scaleMultiplier);
+            }
+        }
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public FloatMask scaleFloatMaskFromTo(float oldFloatMax, float newFloatMax) {
+        float scaleMultiplier = newFloatMax / oldFloatMax;
+        for (int x = 0; x < getSize(); x++) {
```
Sheikah45: same use multiply(scaleMultiplier)

Also not clear on when you would use this over sclaeFloatMask

Sheikah45: or why you wouldnt just use multiply

src/java/map/FloatMask.java
```diff
@@ -572,6 +572,64 @@ public FloatMask removeAreasOfSpecifiedSizeWithLocalMaximums(int minSize, int ma
...
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public int getRankOfPoint(int x, int y) {
+        int rank = 1;
+        float value = get(x, y);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                if(value > get(a, b) || (value == get(a, b) && x * 100 + y > a * 100 + b)) {
```
Sheikah45, line 594: this will result in a lot of ranks due to it being float values. 

also what is the 100 for?

src/java/map/FloatMask.java
```diff
@@ -572,6 +572,64 @@ public FloatMask removeAreasOfSpecifiedSizeWithLocalMaximums(int minSize, int ma
...
+        }
+        return rank;
+    }
+
+    public int getRankOfPointInArea(int x, int y, BinaryMask Area) {
+        int rank = 1;
+        float value = get(x, y);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                if(Area.get(a, b) && (value > get(a, b) || (value == get(a, b) && x * 100 + y > a * 100 + b))) {
```
Sheikah45, line 607: same as above

src/java/map/BinaryMask.java
```diff
@@ -811,6 +849,62 @@ public BinaryMask getAreasWithinSpecifiedDistanceOfEdges(int distance) {
         return newMask;
     }
 
+    public FloatMask getDistanceFieldFromPoint(int x, int y) {
```
Sheikah45, line 837: this should probably be static as it doesnt use any fields.

src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
...
+                        a = getSize();
+                        b = getSize();
+                    }
+                }
+            }
+        }
+        return nearestPoint;
+    }
+
+    public float getLowestDistanceFromPointToArea(int x, int y) {
```
Sheikah45, line 114: This is the same as getDistanceField().get(x,y)

src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
...
+    public float getDistanceBetweenPoints(int x1, int y1, Point Point2) {
+        return (float) StrictMath.sqrt((double) ((x1 - Point2.x) ^ 2 + (y1 - Point2.y) ^ 2));
+    }
+
+    public float getDistanceBetweenPoints(Point Point1, Point Point2) {
+        return (float) StrictMath.sqrt((double) ((Point1.x - Point2.x) ^ 2 + (Point2.x - Point2.y) ^ 2));
+    }
+
+    public Point getNearestPointInAreaTo(int x, int y) {
+        Point nearestPoint = new Point (x, y);
```
Sheikah45: Use Vector2f not Point

src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
...
+        int rank;
+        FloatMask distanceField = getDistanceFieldFromPointInArea(x, y);
```
Sheikah45: Use getDistanceField

src/java/map/BinaryMask.java
```diff
@@ -811,6 +849,62 @@ public BinaryMask getAreasWithinSpecifiedDistanceOfEdges(int distance) {
...
+            for (int b = 0; b < getSize(); b++) {
+                if (this.get(a, b)) {
+                    rankedDistanceField.set(a, b, distanceField.getRankOfPointInArea(a, b, this));
+                }
+            }
+        }
+        return rankedDistanceField;
+    }
+
+    public FloatMask getDistanceFieldFromArea() {
```
Sheikah45, line 883: same as getDistanceField

src/java/map/BinaryMask.java
```diff
@@ -811,6 +849,62 @@ public BinaryMask getAreasWithinSpecifiedDistanceOfEdges(int distance) {
...
+            for (int b = 0; b < getSize(); b++) {
+                if (this.get(a, b)) {
+                    distanceField.set(a, b, getDistanceBetweenPoints(x, y, a, b));
+                }
+            }
+        }
+        return distanceField;
+    }
+
+    public FloatMask getRankedDistanceFieldFromPoint(int x, int y) {
```
Sheikah45, line 859: This is static

src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
...
+
+    public float getDistanceBetweenPoints(Point Point1, Point Point2) {
+        return (float) StrictMath.sqrt((double) ((Point1.x - Point2.x) ^ 2 + (Point2.x - Point2.y) ^ 2));
+    }
+
+    public Point getNearestPointInAreaTo(int x, int y) {
+        Point nearestPoint = new Point (x, y);
+        int rank;
+        FloatMask distanceField = getDistanceFieldFromPointInArea(x, y);
+        for (int a = 0; a < getSize(); a++) {
```
Sheikah45: You should check if the point is n area then just return point if it is. If it is not in area you only need to check the edges of area not the entire mask

src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
...
+            for (int b = 0; b < getSize(); b++) {
+                if (this.get(a, b)) {
+                    rank = distanceField.getRankOfPointInArea(a, b, this);
+                    if(rank == 1) {
+                        nearestPoint.setLocation(a, b);
+                        a = getSize();
```
Sheikah45: if you are trying to get out of the for loop here you could just return now

src/java/map/BinaryMask.java
```diff
@@ -93,6 +96,41 @@ public void set(int x, int y, boolean value) {
...
+    }
+
+    public Point getNearestPointInAreaTo(int x, int y) {
+        Point nearestPoint = new Point (x, y);
+        int rank;
+        FloatMask distanceField = getDistanceFieldFromPointInArea(x, y);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                if (this.get(a, b)) {
+                    rank = distanceField.getRankOfPointInArea(a, b, this);
```
Sheikah45: why use rank instead of keeping track of the minimum distance? I am not sure rank guarantees you the minimum distance

Sheikah45: this will also cause you to loop through the mask again for every point. It would be more efficient to calculate the distance field once and then use it rather than calculate it every time

Sheikah45: if you use getDistanceField then you will get the distance of the closest true point so you will know what radius to check

src/java/map/BinaryMask.java
```diff
@@ -811,6 +849,62 @@ public BinaryMask getAreasWithinSpecifiedDistanceOfEdges(int distance) {
...
+        return rankedDistanceField;
+    }
+
+    public FloatMask getRankedDistanceFieldFromPointInArea(int x, int y) {
+        FloatMask distanceField = getDistanceFieldFromPointInArea(x, y);
+        FloatMask rankedDistanceField = new FloatMask(getSize(), null, symmetrySettings);
+        for (int a = 0; a < getSize(); a++) {
+            for (int b = 0; b < getSize(); b++) {
+                if (this.get(a, b)) {
+                    rankedDistanceField.set(a, b, distanceField.getRankOfPointInArea(a, b, this));
```
Sheikah45, line 876: this will be slow shouldnt need to calculate the rank everytime for everypoint

Sheikah45 (changes requested):
getNearestPointInAreaTo could be much more efficient if you just use the distancefield.

Also I am not a huge fan of the word area as that made me think it was a specified area rather than where the mask is true. I dont think you need the words Area or InArea in the titles.

Also I do not understand why you are using ranks instead of just the distance values

Also you reimplemented some functions

src/java/map/FloatMask.java
Penguin5: generic large multiplier to make it so it prioritizes x-values before y-values for tie-breaking (I did this as an alternative to more if-statements for the same functionality)

src/java/map/FloatMask.java
Penguin5: same response

src/java/map/BinaryMask.java
Penguin5: it gets the size and such from the mask, so I think it makes sense to keep? - open to alternate suggestions though...

src/java/map/BinaryMask.java
Penguin5: function has been changed - I could do getDistanceField().get(x, y) instead if you think that would be more efficient than what I have now?

src/java/map/BinaryMask.java
Penguin5: Now that things have been adjusted; how do you think the efficiency compares?  Should I just delete this?

src/java/map/BinaryMask.java
Penguin5: is there a problem?

src/java/map/BinaryMask.java
Penguin5: Keeping track of the minimum distance won't account for ties?
I also want to be able to get areas of set sizes with this.
I am open to suggestions that improve efficiency but don't reduce quality here.

src/java/map/BinaryMask.java
Penguin5: I'm open to better suggestions that don't remove the intended functionality.

src/java/map/BinaryMask.java
Sheikah45: this will be fine and more efficient than that I think

src/java/map/BinaryMask.java
Sheikah45: ah I missed that getSize call

Sheikah45: although it could just take that as a parameter but doesnt matter much

src/java/map/BinaryMask.java
Sheikah45: yeah this is still going to be much slower than getDistanceField. That is because for every point it is going to iterate through every edge point so the speed will also vary based on the number of edge points so it will be more cumbersome where as getDistanceField calculates it all at once without all the point to point distance calculations

src/java/map/FloatMask.java
Sheikah45: would it be better to make the multiplier the size? Same effect but is also related to the mask

Sheikah45:
I tried testing this and got errors because getDistanceBetweenPoints no longer exists in  BinaryMask

PR #247, opened 2020-12-01
Redid generator setupHeightmapPipeline and added new brush usage

Penguin5:
Created ConcurrentFloatMask functions: clampMaxInArea, clampMinInArea, useBrush, useBrushRepeatedlyCenteredWithinArea, and useBrushRepeatedlyCenteredWithinAreaToDensity
Created FloatMask functions: clampMinInArea, clampMaxInArea, addFloatMaskCenteredAtLocationWithSize, useBrush, useBrushRepeatedlyCenteredWithinArea, and useBrushRepeatedlyCenteredWithinAreaToDensity

Generator-specific changes:
Redid terrain construction, primarily with new brush functions
Replaced much of the setupHeightmapPipeline
Modified/lessened ramp mask placement
Created some height variables
Changed texturing to be somewhat closer to populator texturing (more precise than previous generator texturing)
Forced flat player spawn areas after most terrain generation/changes

### DIFF
--- a/src/java/generator/MapGenerator.java
+++ b/src/java/generator/MapGenerator.java
@@ -46,9 +46,10 @@ public strictfp class MapGenerator {
     public static final float PLATEAU_DENSITY_MIN = .35f;
     public static final float PLATEAU_DENSITY_MAX = .5f;
     public static final float PLATEAU_DENSITY_RANGE = PLATEAU_DENSITY_MAX - PLATEAU_DENSITY_MIN;
-    public static final float PLATEAU_HEIGHT = 5f;
-    public static final float VALLEY_HEIGHT = -2f;
-    public static final float HILL_HEIGHT = 2f;
+    public static final float PLATEAU_HEIGHT = 7f;
+    public static final float OCEAN_FLOOR = -15f;
+    public static final float VALLEY_FLOOR = -5f;
+    public static final float LAND_HEIGHT = .5f;
     public static boolean DEBUG = false;
     //read from cli args
     private String pathToFolder = ".";
@@ -826,24 +827,19 @@ public strictfp class MapGenerator {
         String brush3 = Brushes.goodBrushes.get(random.nextInt(length));
         String brush4 = Brushes.goodBrushes.get(random.nextInt(length));
 
-        float deepWaterFloorHeight = -15f;
-        float plateauHeight = 7f;
-        float plateauValleyFloorHeight = -5f;
-        float landFloorHeight = .5f;
-
         heightmapMountains.useBrushRepeatedlyCenteredWithinAreaToDensity(mountains.deflate(2), brush3, 32, .30f, 5f).smooth(1);
 
-        ConcurrentBinaryMask paintedMountains = new ConcurrentBinaryMask(heightmapMountains, plateauHeight / 2, random.nextLong(), "paintedMountains");
+        ConcurrentBinaryMask paintedMountains = new ConcurrentBinaryMask(heightmapMountains, PLATEAU_HEIGHT / 2, random.nextLong(), "paintedMountains");
 
         plateaus.combine(paintedMountains.intersect(plateaus.copy().inflate(32)));
 
-        heightmapPlateaus.useBrushRepeatedlyCenteredWithinAreaToDensity(plateaus.deflate(8), brush1, 42, .16f, 14f).clampMax(plateauHeight);
+        heightmapPlateaus.useBrushRepeatedlyCenteredWithinAreaToDensity(plateaus.deflate(8), brush1, 42, .16f, 14f).clampMax(PLATEAU_HEIGHT);
         heightmapValleys.useBrushRepeatedlyCenteredWithinAreaToDensity(valleys, brush2, 24, .28f, -0.25f)
-                .clampMin(plateauValleyFloorHeight).smooth(2);
+                .clampMin(VALLEY_FLOOR).smooth(2);
         heightmapHills.useBrushRepeatedlyCenteredWithinAreaToDensity(hills, brush4, 24, .28f, 0.25f)
                 .smooth(2);
 
-        ConcurrentBinaryMask paintedPlateaus = new ConcurrentBinaryMask(heightmapPlateaus, plateauHeight / 2, random.nextLong(), "paintedPlateaus");
+        ConcurrentBinaryMask paintedPlateaus = new ConcurrentBinaryMask(heightmapPlateaus, PLATEAU_HEIGHT / 2, random.nextLong(), "paintedPlateaus");
 
         ConcurrentBinaryMask spawnRamps = new ConcurrentBinaryMask(mapSize + 1, random.nextLong(), symmetrySettings, "spawnRamps");
         ramps.combine(paintedPlateaus.copy().outline()).minus(paintedMountains.copy().inflate(8)).flipValues(rampDensity * .004f + .002f).inflate(12);
@@ -852,15 +848,15 @@ public strictfp class MapGenerator {
 
         land.combine(paintedPlateaus);
 
-        heightmapBase.addDistance(land, -.5f).clampMin(deepWaterFloorHeight).smooth(4, land.copy().invert().deflate(4));
+        heightmapBase.addDistance(land, -.5f).clampMin(OCEAN_FLOOR).smooth(4, land.copy().invert().deflate(4));
 
-        heightmapLand.add(heightmapHills).add(heightmapValleys).add(heightmapMountains).setValueInArea(landFloorHeight, spawnLandMask).add(heightmapPlateaus).setValueInArea(plateauHeight, spawnPlateauMask);
+        heightmapLand.add(heightmapHills).add(heightmapValleys).add(heightmapMountains).setValueInArea(LAND_HEIGHT, spawnLandMask).add(heightmapPlateaus).setValueInArea(PLATEAU_HEIGHT, spawnPlateauMask);
         heightmapLand.smooth(24, ramps).smooth(16, ramps.copy().inflate(10)).smooth(4, ramps.copy().inflate(14)).smooth(2, ramps.copy().inflate(18));
 
         heightmapBase.add(heightmapLand);
         heightmapBase.smooth(4, spawnLandMask.copy().combine(spawnPlateauMask).inflate(4));
 
-        heightmapBase.addAll(waterHeight + landFloorHeight).addGaussianNoise(.025f).clampMin(0f).clampMax(256f);
+        heightmapBase.addAll(waterHeight + LAND_HEIGHT).addGaussianNoise(.025f).clampMin(0f).clampMax(256f);
 
         slope = heightmapBase.copy().supcomGradient();
 
--- a/src/java/generator/MapGenerator.java
+++ b/src/java/generator/MapGenerator.java
@@ -2,6 +2,7 @@ package generator;
 
 import biomes.Biome;
 import biomes.Biomes;
+import brushes.Brushes;
 import com.google.common.io.BaseEncoding;
 import exporter.MapExporter;
 import exporter.SCMapExporter;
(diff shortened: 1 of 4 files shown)


src/java/generator/MapGenerator.java
```diff
@@ -834,31 +835,37 @@ private void setupTerrainPipeline() {
...
-
-        heightmapMountains.add(mountainsPresent.copy().minus(plateaus), PLATEAU_HEIGHT).subtract(mountainsPresent.copy().invert(), 10f).clampMin(0);
-
-        heightmapBase.add(heightmapMountains).add(heightmapLand).add(heightmapPlateaus).smooth(1);
+        ConcurrentBinaryMask land2 = land.copy().setSize(mapSize + 1);
+
+        List<String> allBrushes = Arrays.asList("buttons.png", "crescent.png", "crystal.png", "hawaii1.png", "hawaii2.png", "hill1.png", "hill2.png", "mountain1.png", "mountain2.png", "mountain3.png", "mountain4.png", "mountain5.png", "mountain6.png", "mountain7.png", "mountain8.png", "mountain9.png", "noise1.png", "noise2.png", "striations.png", "structured.png", "sun.png", "volcano2.png");
+        List<String> okayBrushes = Arrays.asList("hawaii1.png", "hawaii2.png", "hill1.png", "hill2.png", "mountain1.png", "mountain2.png", "mountain3.png", "mountain4.png", "mountain5.png", "mountain6.png", "mountain7.png", "mountain8.png", "structured.png", "volcano2.png");
+        List<String> goodBrushes = Arrays.asList("hawaii2.png", "hill1.png", "mountain1.png", "mountain2.png", "mountain4.png", "mountain5.png", "mountain6.png", "mountain7.png", "volcano2.png");
```
Sheikah45: These should probably be static lists in the Brushes class and not defined in the generator class. Also if only goodBrushes is used better to only have that one defined.

src/java/generator/MapGenerator.java
```diff
@@ -834,31 +835,37 @@ private void setupTerrainPipeline() {
...
-        heightmapCliffs.init(cliffs, 0, 1f).maskToMountains(cliffs);
-        heightmapShore.init(shore, 0, 1.5f).maskToMountains(shore);
-        heightmapMountains.maskToMountains(mountains).add(heightmapCliffs).add(heightmapShore);
-
-        ConcurrentBinaryMask mountainsPresent = new ConcurrentBinaryMask(heightmapMountains, 3f, random.nextLong(), "mountainsPresent");
-
-        heightmapMountains.add(mountainsPresent.copy().minus(plateaus), PLATEAU_HEIGHT).subtract(mountainsPresent.copy().invert(), 10f).clampMin(0);
-
-        heightmapBase.add(heightmapMountains).add(heightmapLand).add(heightmapPlateaus).smooth(1);
+        ConcurrentBinaryMask land2 = land.copy().setSize(mapSize + 1);
```
Sheikah45: Why do you need this copy of land it looks like you just make copies of it anyway and land should already be mapSize + 1 at this point

src/java/generator/MapGenerator.java
```diff
@@ -834,31 +835,37 @@ private void setupTerrainPipeline() {
...
+
+        float deepWaterFloorHeight = waterHeight - 15f;
+        float shallowWaterFloorHeight = waterHeight - 8f;
+        float landFloorHeight = waterHeight + 0.5f;
+        float landValleyFloorHeight = waterHeight + 0.1f;
+        float plateauHeight = waterHeight + 7f;
+        float plateauValleyFloorHeight = plateauHeight - 5f;
+
+        heightmapBase.clampMin(deepWaterFloorHeight).clampMinInArea(shallowWaterFloorHeight, land2.copy().inflate(23)).clampMinInArea(landFloorHeight, land2).smooth(30).clampMinInArea(landFloorHeight, land2.copy().inflate(8)).smooth(8);
+        heightmapBase.useBrushRepeatedlyCenteredWithinAreaToDensity(plateaus, brush1, 42,16,14f).clampMax(plateauHeight);
```
Sheikah45: spaces in between commas

src/java/map/FloatMask.java
```diff
@@ -423,6 +420,21 @@ public FloatMask clampMin(float val) {
         return this;
     }
 
+    public FloatMask clampMinInArea(float val, BinaryMask area) {
+        if (area.getSize() != getSize()) {
+            throw new IllegalArgumentException("Masks not the same size: area is " + area.getSize() + " and FloatMask is " + getSize());
+        }
+        for (int y = 0; y < getSize(); y++) {
+            for (int x = 0; x < getSize(); x++) {
+                if(area.get(x, y)) {
```
Sheikah45: space between if and parentheses

src/java/map/FloatMask.java
```diff
@@ -458,6 +470,21 @@ public FloatMask clampMax(float val) {
         return this;
     }
 
+    public FloatMask clampMaxInArea(float val, BinaryMask area) {
+        if (area.getSize() != getSize()) {
+            throw new IllegalArgumentException("Masks not the same size: area is " + area.getSize() + " and FloatMask is " + getSize());
+        }
+        for (int y = 0; y < getSize(); y++) {
+            for (int x = 0; x < getSize(); x++) {
+                if(area.get(x, y)) {
```
Sheikah45: whitespace

src/java/map/FloatMask.java
```diff
@@ -761,6 +788,48 @@ public FloatMask spike(int radius, BinaryMask limiter) {
...
+    public FloatMask useBrushRepeatedlyCenteredWithinArea(BinaryMask area, String brushName, int size, int frequency, float intensity) {
+        if (size > getSize()) {
+            throw new IllegalArgumentException("Added mask size is larger than base mask size");
+        }
+        List<Vector2f> possibleLocations = new ArrayList<Vector2f>(area.getAllCoordinatesEqualTo(true, 1));
+        int length = possibleLocations.size();
+        FloatMask brush = loadBrush(brushName, new SymmetrySettings(Symmetry.NONE, Symmetry.NONE, Symmetry.NONE));
+        brush.multiply(intensity / brush.getMax()).setSize(size);
+        for (int z = 0; z < frequency; z++) {
+            add(brush, possibleLocations.get(random.nextInt(length)), true);
```
Sheikah45: If you dont want it to repeat positions you can use remove instead of get. You would then need to check if any are left as well. I would suggest using a Linked list then

src/java/map/FloatMask.java
```diff
@@ -761,6 +788,48 @@ public FloatMask spike(int radius, BinaryMask limiter) {
...
+        }
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
+
+    public FloatMask useBrushRepeatedlyCenteredWithinAreaToDensity(BinaryMask area, String brushName, int size, float density, float intensity) {
+        List<Vector2f> possibleLocations = new ArrayList<Vector2f>(area.getAllCoordinatesEqualTo(true, 1));
```
Sheikah45: useBurshRepeatedlyCenteredWithinArea also gets all the possible Locations so this call or that one should be removed for efficiency as they are rather expensive

src/java/map/FloatMask.java
```diff
@@ -761,6 +788,48 @@ public FloatMask spike(int radius, BinaryMask limiter) {
...
+        int length = possibleLocations.size();
+        int frequency = (int) (density * (float) length / (float) 2621);
```
Sheikah45: what is with the arbitrary number? How does this then scale with mask size?

Sheikah45: Also you can add f to the end of a number to make it a float rather than casting it

src/java/map/ConcurrentFloatMask.java
```diff
@@ -97,18 +98,48 @@ public ConcurrentFloatMask clampMax(float value) {
...
+        return Pipeline.add(this, Collections.singletonList(this), res ->
+                this.floatMask.useBrushRepeatedlyCenteredWithinArea(area.getBinaryMask(), brushName, size, frequency, intensity)
+        );
+    }
+
+    public ConcurrentFloatMask useBrushRepeatedlyCenteredWithinAreaToDensity(ConcurrentBinaryMask area, String brushName, int size, float density, float intensity) {
+        return Pipeline.add(this, Collections.singletonList(this), res ->
+                this.floatMask.useBrushRepeatedlyCenteredWithinAreaToDensity(area.getBinaryMask(), brushName, size, density, intensity)
+        );
+    }
```
Sheikah45: This has the wrong dependencies list as area is not included. Also it needs to be res.get(1) not area in the function

src/java/map/ConcurrentFloatMask.java
```diff
@@ -97,18 +98,48 @@ public ConcurrentFloatMask clampMax(float value) {
...
+        return Pipeline.add(this, Collections.singletonList(this), res ->
+                this.floatMask.useBrush(location, brushName, intensity, size)
+        );
+    }
+
+    public ConcurrentFloatMask useBrushRepeatedlyCenteredWithinArea(ConcurrentBinaryMask area, String brushName, int size, int frequency, float intensity) {
+        return Pipeline.add(this, Collections.singletonList(this), res ->
+                this.floatMask.useBrushRepeatedlyCenteredWithinArea(area.getBinaryMask(), brushName, size, frequency, intensity)
+        );
+    }
```
Sheikah45: This has the wrong dependencies list as area is not included. Also it needs to be res.get(1) not area in the function

src/java/map/ConcurrentFloatMask.java
```diff
@@ -97,18 +98,48 @@ public ConcurrentFloatMask clampMax(float value) {
...
         return Pipeline.add(this, Collections.singletonList(this), res ->
                 this.floatMask.threshold(value)
         );
     }
 
+    public ConcurrentFloatMask useBrush(Vector2f location, String brushName, float intensity, int size) {
+        return Pipeline.add(this, Collections.singletonList(this), res ->
+                this.floatMask.useBrush(location, brushName, intensity, size)
+        );
+    }
```
Sheikah45: This function most likely wont work with concurrent masks if it needs a location from another mask as the location wont be defined when this is called.

src/java/generator/MapGenerator.java
```diff
@@ -790,6 +790,7 @@ private void setupTerrainPipeline() {
         spawnRamps.combine(spawnLandMask.copy().outline()).combine(spawnPlateauMask.copy().outline()).inflate(24).intersect(plateaus.copy().outline()).flipValues(.0075f).inflate(24);
 
         ramps.combine(spawnRamps).fillGaps(48).smooth(8, .25f);
+        ramps.acid(0.00034f, 35f).deflate(5f).inflate(15f).smooth(8).inflate(4);
```
Sheikah45: Might be better to just lower the initial density of the ramps

Sheikah45 (changes requested):

src/java/generator/MapGenerator.java
Penguin5: moved

src/java/generator/MapGenerator.java
Penguin5: regular land gave errors at least some times when I used it with setting the size
and I do use land2 in multiple cases (not just with copying it)

src/java/map/FloatMask.java
Penguin5: I do want it to repeat...

src/java/map/FloatMask.java
Penguin5: The number is approximately the area of 1 square km.  I thoroughly evaluated changing it with size and tested using that, and I determined this way seemed superior.
Changed to use f

src/java/map/FloatMask.java
Penguin5: changed

src/java/map/ConcurrentFloatMask.java
Penguin5: fixed

src/java/map/ConcurrentFloatMask.java
Penguin5: fixed

Sheikah45 (approved):

PR #257, opened 2020-12-31
Created FloatMask to PNG writing functionality, MapImageWriter class, ImageGenerator class, and related functions

Penguin5:


src/java/mapImageWriter/MapImageWriter.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    private void interpretArguments(Map<String, String> arguments) {
+        if (arguments.containsKey("help")) {
+            System.out.println("map-transformer usage:\n" +
+                    "--help                 produce help message\n" +
+                    "--in-folder-path arg   required, set the input folder for the map - the images will appear in this folder\n" +
+                    "--team-symmetry arg    required, set the symmetry for the teams(X, Z, XZ, ZX)\n" +
+                    "--spawn-symmetry arg   required, set the symmetry for the spawns(POINT, X, Z, XZ, ZX)\n" +
+                    "--texture-res arg      optional, set arg texture resolution (128, 256, 512, 1024, 2048, etc) - resolution cannot exceed map size (256 = 5 km)\n" +
```
Sheikah45: These options are not required or even needed as you could just set the symmetry to NONE since you are just copying the masks and writing to png

src/java/mapImageWriter/MapImageWriter.java
```diff
@@ -0,0 +1,192 @@
...
+        mapImageWriter.writeMapImages();
+    }
+
+    public void interpretArguments(String[] args) {
+        interpretArguments(ArgumentParser.parse(args));
+    }
+
+    private void interpretArguments(Map<String, String> arguments) {
+        if (arguments.containsKey("help")) {
+            System.out.println("map-transformer usage:\n" +
```
Sheikah45: This isnt the map transformer

src/java/mapImageWriter/MapImageWriter.java
```diff
@@ -0,0 +1,192 @@
...
+                    "--help                 produce help message\n" +
+                    "--in-folder-path arg   required, set the input folder for the map - the images will appear in this folder\n" +
+                    "--team-symmetry arg    required, set the symmetry for the teams(X, Z, XZ, ZX)\n" +
+                    "--spawn-symmetry arg   required, set the symmetry for the spawns(POINT, X, Z, XZ, ZX)\n" +
+                    "--texture-res arg      optional, set arg texture resolution (128, 256, 512, 1024, 2048, etc) - resolution cannot exceed map size (256 = 5 km)\n" +
+                    "--create-images arg    optional, create images arg determines which layers have PNG images created from them (0, 1, 2, 3, 4, 5, 6, 7, 8, h)\n" +
+                    " - ie: to create all available PNG images except the one for layer 7, use: --create-images 01234568h\n" +
+                    " - options 0 - 8 will create PNG's of the corresponding texture layers, and option h will create a PNG of the heightmap" +
```
Sheikah45: This argument is a little unneeded as all this class does is create images it could be an optional parameter like layers

src/java/mapImageWriter/MapImageWriter.java
```diff
@@ -0,0 +1,192 @@
...
+
+    public void writeMapImages() throws IOException {
+
+        Random random = new Random();
+        FloatMask heightmapBase = map.getHeightMask(symmetrySettings);
+        heightmapBase = new FloatMask(heightmapBase, random.nextLong());
+        heightmapBase.applySymmetry(SymmetryType.SPAWN);
+        map.setHeightImage(heightmapBase);
+
+        if(writeImages) {
```
Sheikah45: This is unneeded as if you didnt want to write images you wouldnt have run this

src/java/mapImageWriter/MapImageWriter.java
Penguin5: lol, I'll fix the populator labelling too then xD

src/java/map/BinaryMask.java
```diff
@@ -167,6 +167,21 @@ public BinaryMask randomWalkWithBrush(Vector2f start, String brushName, int size
         return this;
     }
 
+    public BinaryMask guidedWalkWithBrush(Vector2f start, Vector2f target, String brushName, int size, int numberOfUses,
```
Sheikah45, line 170: All the other connect with brush functions should use this to reduce code duplication

src/java/map/FloatMask.java
```diff
@@ -801,6 +801,25 @@ public FloatMask useBrushWithinAreaWithDensity(BinaryMask area, String brushName
...
+        }
+        for (int x = 0; x < size; x += size - 1) {
+            for (int y = 0; y < size; y++) {
+                if(getValueAt(x, y) > value) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
```
Sheikah45: Look at BinaryMask.fillEdge for better loop construction to check the edge

src/java/util/ImageUtils.java
```diff
@@ -45,4 +49,48 @@ public static BufferedImage scaleImage(BufferedImage image, int width, int heigh
         return imageScaled;
     }
 
+    public static void writePNGFromMasks(FloatMask redMask, FloatMask greenMask, FloatMask blueMask, float scaleMultiplier, String path) throws IOException {
```
Sheikah45: make path variable a Path object for better cross platform usage

src/java/mapImageWriter/MapImageWriter.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+        if(writeLayer6) {
+            util.ImageUtils.writePNGFromMask(oldLayer6, 255, writeImagesPath + "\\Layer 6.png");
+        }
+        if(writeLayer7) {
+            util.ImageUtils.writePNGFromMask(oldLayer7, 255, writeImagesPath + "\\Layer 7.png");
+        }
+        if(writeLayer8) {
+            util.ImageUtils.writePNGFromMask(oldLayer8, 255, writeImagesPath + "\\Layer 8.png");
+        }
```
Sheikah45: make writeImagesPath a Path object and use resolve for better cross platform

src/java/map/BinaryMask.java
Penguin5: It's not the same thing though.  This one is guided, while other connect with brush functions are not guided like this.  There are times when non-guided connect functions are better to use.

src/java/mapImageWriter/MapImageWriter.java
Penguin5: I used Paths.get() instead

src/java/imageGenerator/ImageGenerator.java
```diff
@@ -0,0 +1,275 @@
+package imageGenerator;
```
Sheikah45: I would just move this into the generator package

src/java/imageGenerator/ImageGenerator.java
```diff
@@ -0,0 +1,275 @@
...
+            base.combineBrush(new Vector2f(center + random.nextInt(variationDistance) - random.nextInt(variationDistance), center + random.nextInt(variationDistance) - random.nextInt(variationDistance)), brush2, random.nextFloat(), 1f, reducedSize);
+            base.combineBrush(new Vector2f(center + random.nextInt(variationDistance) - random.nextInt(variationDistance), center + random.nextInt(variationDistance) - random.nextInt(variationDistance)), brush3, random.nextFloat(), 1f, reducedSize);
+            base.combineBrush(new Vector2f(center + random.nextInt(variationDistance) - random.nextInt(variationDistance), center + random.nextInt(variationDistance) - random.nextInt(variationDistance)), brush4, random.nextFloat(), 1f, reducedSize);
+            base.combineBrush(new Vector2f(center + random.nextInt(variationDistance) - random.nextInt(variationDistance), center + random.nextInt(variationDistance) - random.nextInt(variationDistance)), brush5, random.nextFloat(), 1f, reducedSize);
```
Sheikah45: I would just break this into multiline

src/java/imageGenerator/ImageGenerator.java
```diff
@@ -0,0 +1,275 @@
...
+                util.ImageUtils.writeAutoScaledPNGFromMask(newBrush, Paths.get(folderPath + "\\Brush_" + (i + 1) + ".png"));
+            }
+        }
+    }
+
+    public void generateCustomTextures(int size, int numberToGenerate, float colorVariation) throws IOException {
+
+        float redLocus;
+        float greenLocus;
+        float blueLocus;
```
Sheikah45: Locus?

src/java/imageGenerator/ImageGenerator.java
```diff
@@ -0,0 +1,275 @@
...
+                    if(areaToTexture.getCount() > size * 3) {
+                        tooEmpty = false;
+                        areaToTexture = wholeImage;
+                    }
+                }
+                for (int x = 0; x < 5; x++) {
+                    Vector2f loc = areaToTexture.getRandomPosition();
+                    while(loc == null) {
+                        loc = wholeImage.getRandomPosition();
+                    }
```
Sheikah45: if loc == null then there are no more places that are true so this will be an infinite loop

src/java/imageGenerator/ImageGenerator.java
```diff
@@ -0,0 +1,275 @@
...
+                    Vector2f target = areaToTexture.getRandomPosition();
+                    while(target == null) {
+                        target = wholeImage.getRandomPosition();
+                    }
```
Sheikah45: same infinite loop condition

src/java/imageGenerator/ImageGenerator.java
Sheikah45: The getRandomPosition is also not as efficient you could use getRandomCoordinatesEqualTo which returns a list of coordinates where it is true in random order. Then could use this.

src/java/imageGenerator/ImageGenerator.java
```diff
@@ -0,0 +1,275 @@
...
+                    chain.guidedWalkWithBrushToroidally(loc, target, Brushes.goodBrushes.get(random.nextInt(brushListLength)), chainBrushSize,
```
Sheikah45: Toroidally?

src/java/imageGenerator/ImageGenerator.java
```diff
@@ -0,0 +1,275 @@
...
+                            random.nextInt(15) + 1, 0.1f, 1f, chainBrushSize / 2);
+                }
+                chainTexture.useBrushWithinAreaWithDensityTorroidally(chain, Brushes.goodBrushes.get(random.nextInt(brushListLength)), chainTextureBrushSize, 0.05f, 5 * random.nextFloat());
+
+                float redWeight = redLocus + (random.nextBoolean() ? random.nextFloat() * colorVariation / 100 : - random.nextFloat() * colorVariation / 100);
+                float greenWeight = greenLocus + (random.nextBoolean() ? random.nextFloat() * colorVariation / 100 : - random.nextFloat() * colorVariation / 100);
+                float blueWeight = blueLocus + (random.nextBoolean() ? random.nextFloat() * colorVariation / 100 : - random.nextFloat() * colorVariation / 100);
```
Sheikah45: (random.nextBoolean() ? 1 : -1) * random.nextFloat() * colorVariation / 100

src/java/map/BinaryMask.java
Sheikah45: How is this one guided it seems like it is using the same inner logic just it is a specific start and target rather than chosen randomly from a list

src/java/map/FloatMask.java
```diff
@@ -261,6 +261,63 @@ public FloatMask add(float val) {
...
+    public FloatMask addWeighted(FloatMask other, float weight) {
+        checkSize(other);
+        for (int y = 0; y < getSize(); y++) {
+            for (int x = 0; x < getSize(); x++) {
+                addValueAt(x, y, other.getValueAt(x, y) * weight);
+            }
+        }
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
```
Sheikah45, line 273: might be better called scale

Sheikah45: also could use the multiply function

src/java/map/FloatMask.java
```diff
@@ -261,6 +261,63 @@ public FloatMask add(float val) {
...
+
+    public FloatMask addTorroidally(FloatMask other) {
```
Sheikah45: Sometimes Torroidally has 2 rs and other times only 2

src/java/map/FloatMask.java
```diff
@@ -261,6 +261,63 @@ public FloatMask add(float val) {
...
+        int size = getSize();
+        int otherSize = other.getSize();
+        for (int y = 0; y < otherSize; y++) {
+            for (int x = 0; x < otherSize; x++) {
+                addValueAt(x % size, y % size, other.getValueAt(x, y) * weight);
+            }
+        }
+        VisualDebugger.visualizeMask(this);
+        return this;
+    }
```
Sheikah45: same weight vs scale

src/java/mapImageWriter/MapImageWriter.java
```diff
@@ -0,0 +1,190 @@
+package mapImageWriter;
```
Sheikah45: might be better in the exported package

Sheikah45 (changes requested):
Should the toroidally functions assume the other mask is larger?

src/java/imageGenerator/ImageGenerator.java
Penguin5: "locus
n.	A locality; a place.
n.	A center or focus of great activity or intense concentration."

IE: The red locus is the focal point upon which the red color weight are chosen (if there is a low color variation value, then all of the red color weight will be close to the value of the red locus.)

src/java/imageGenerator/ImageGenerator.java
Penguin5: Considering the whole image is the basis for that, there either wouldn't be an image to generate (which would make this irrelevant), or there would be some valid positions...

src/java/imageGenerator/ImageGenerator.java
Penguin5: It is like a toroidal shape in regards to how it applies functionality to the mask.  You can imagine the left and right sides being connected, and you can also imagine the top and bottom sides being connected.  If turned into a 3-dimensional representation, it theoretically looks somewhat like a doughnut.

src/java/map/FloatMask.java
Penguin5: Those don't seem like fixes or improvements to me... if you really want to change those, you can...

src/java/map/FloatMask.java
Penguin5: same thing

src/java/mapImageWriter/MapImageWriter.java
Penguin5: okay

src/java/map/FloatMask.java
Penguin5: fixed

Penguin5:
Created new pr

PR #262, opened 2021-01-11
Map resizing

Penguin5:
Added Map Resizing
Created SCMap reSize function
Added Map Resizing/relocating functionality to Map Transformer
Created ImageUtils insertImageIntoNewImageOfSize function
Made Decal scale not final
Made SCMap size not final
Added shrink2 (again), enlarge2 (again), and setSize2 Floatmask functions


src/java/neroxis/map/SCMap.java
```diff
@@ -376,6 +382,257 @@ public void addCubeMap(CubeMap cubeMap) {
...
+        layer5.addWithOffset(oldLayer5, locToPutCenterOfCurrentMapContent, true);
+        FloatMask layer6 =  new FloatMask(newMapBoundsSize, new Random().nextLong(), symmetrySettings);
+        layer6.addWithOffset(oldLayer6, locToPutCenterOfCurrentMapContent, true);
+        FloatMask layer7 =  new FloatMask(newMapBoundsSize, new Random().nextLong(), symmetrySettings);
+        layer7.addWithOffset(oldLayer7, locToPutCenterOfCurrentMapContent, true);
+        FloatMask layer8 =  new FloatMask(newMapBoundsSize, new Random().nextLong(), symmetrySettings);
+        layer8.addWithOffset(oldLayer8, locToPutCenterOfCurrentMapContent, true);
+
+        setTextureMasksLowScaled(layer1, layer2, layer3, layer4);
+        setTextureMasksHighScaled(layer5, layer6, layer7, layer8);
```
Sheikah45: Why are the textureMasks handled any differently they are just bufferedImages as well

src/java/neroxis/map/SCMap.java
```diff
@@ -376,6 +382,257 @@ public void addCubeMap(CubeMap cubeMap) {
         cubeMaps.add(cubeMap);
     }
 
+    public void reSize(int resizeCurrentMapContentTo, int newMapBoundsSize, Vector2f locToPutCenterOfCurrentMapContent) {
```
Sheikah45: resize is one word no camelCase

src/java/neroxis/map/SCMap.java
```diff
@@ -376,6 +382,257 @@ public void addCubeMap(CubeMap cubeMap) {
...
+                    float y;
+                    if(heightmapBase.inBounds((int) x,  (int) z)) {
+                        y = heightmapBase.getValueAt((int) x, (int) z);
+                    } else {
+                        y = 0; }
+                    group.getUnit(b).setPosition(new Vector3f(x, y, z));
+                }
+            }
+        }
+    }
```
Sheikah45: A function should be made that changes the position as this is all the same operation and is a lot of duplicated code. Function could take in the list of locations or there could be a class each of these inherits from that could be used.

Sheikah45: Also would want to use a stream for each rather than a for loop

src/java/neroxis/transformer/MapTransformer.java
```diff
@@ -37,6 +37,11 @@
     private boolean reverseSide;
     private float angle;
     private SymmetrySettings symmetrySettings;
+    private int resize = -1;
+    private int mapSize = -1;
+    private int oldMapSize;
+    private int shiftX = 1000000;
+    private int shiftZ = 1000000;
```
Sheikah45: If you are just going to default to mapSize / 2 there is no need to set the value here. Just set it using an else in the argument parsing.

src/java/neroxis/transformer/MapTransformer.java
```diff
@@ -37,6 +37,11 @@
     private boolean reverseSide;
     private float angle;
     private SymmetrySettings symmetrySettings;
+    private int resize = -1;
+    private int mapSize = -1;
```
Sheikah45: If you are just going to default the values there is no need to set the value here. Just set it using an else in the argument parsing.

src/java/neroxis/transformer/MapTransformer.java
```diff
@@ -308,6 +333,24 @@ public void transform() {
...
+        if (mapSize == -1) {
+            mapSize = oldMapSize;
+        }
+        float sizeRatio = (float) resize / (float) oldMapSize;
+        if (shiftX == 1000000) {
+            shiftX = mapSize / 2;
+        }
+        if (shiftZ == 1000000) {
+            shiftZ = mapSize / 2;
+        }
```
Sheikah45: set these in argument parsing

Sheikah45 (changes requested):

src/java/neroxis/map/SCMap.java
Penguin5: I tried handling them like the others first, but the result of that was that when the texture masks were used to be only part of the new image, the new texture masks were in the bottom right corner of the image rather than in the desired location.

src/java/neroxis/map/SCMap.java
Penguin5: okay

src/java/neroxis/transformer/MapTransformer.java
Emperor-Penguin1: okay

Penguin5: The proper mapSize isn't necessarily set at that point though.

src/java/neroxis/transformer/MapTransformer.java
Penguin5: The proper size can't be defaulted to at that point though.

src/java/neroxis/transformer/MapTransformer.java
Penguin5: I would do that if the size was known definitively at that point.

src/java/neroxis/transformer/MapTransformer.java
Penguin5: made flags

src/java/neroxis/transformer/MapTransformer.java
Penguin5: made flags

src/java/neroxis/transformer/MapTransformer.java
Penguin5: made flags

src/java/neroxis/map/SCMap.java
Sheikah45: i suspect then that is because the texture mask size was different from the map size. As it is done currently the size could still be off as you are assuming the size of the texture masks was the same size as the map which is not always the case. Also I don;t think the scaled float masks should be used as they can cause changes in the textures due to scaling issues. Ultimately better to figure out the sizing issues with the imageutils

Penguin5: I had spent a while trying to figure this issue out better for the populator, and this is the best system for this that we implemented.  It does work consistently, although some gpg textures can be a bit rough around the edges... if you know how to do a better solution for this, feel free to implement it.

src/java/neroxis/map/SCMap.java
```diff
@@ -376,6 +382,266 @@ public void addCubeMap(CubeMap cubeMap) {
         cubeMaps.add(cubeMap);
     }
 
+    public void resize(int resizeCurrentMapContentTo, int newMapBoundsSize, Vector2f locToPutCenterOfCurrentMapContent, float heightMultiplier) {
```
Sheikah45: this function is doing too much now resizing has nothing to do with height multiplication

Penguin5: Resizing has a lot to do with height multiplication.
If you resize a map and you don't multiply the height accordingly, then the slopes will be wrong.  For example, if you take a regular 20 x 20 map and resize it to 5 x 5, you will get many impassable areas that should be passable, unless you scale/multiply the height accordingly.  Conversely, if you take a small map and resize it to be bigger, you will get gentler slopes (which can make previously impassable terrain instead be passable), unless you scale/multiply the height accordingly.

src/java/neroxis/map/SCMap.java
```diff
@@ -54,7 +59,7 @@
     private final ArrayList<AIMarker> expansionAIMarkers;
     private final ArrayList<AIMarker> largeExpansionAIMarkers;
     private final ArrayList<AIMarker> navalAreaAIMarkers;
-    private final ArrayList<AIMarker> navalRallyMarkers;
+    private final ArrayList<AIMarker> NavyRallyMarkers;
```
Sheikah45: variable names start with a lowercase letter

src/java/neroxis/map/SCMap.java
```diff
@@ -100,7 +105,7 @@ public SCMap(int size, int spawnCount, int mexCount, int hydroCount, Biome biome
         expansionAIMarkers = new ArrayList<>();
         largeExpansionAIMarkers = new ArrayList<>();
         navalAreaAIMarkers = new ArrayList<>();
-        navalRallyMarkers = new ArrayList<>();
+        NavyRallyMarkers = new ArrayList<>();
```
Sheikah45: lowercase

src/java/neroxis/map/SCMap.java
```diff
@@ -341,15 +346,15 @@ public void addNavalAreaMarker(AIMarker aiMarker) {
     }
 
     public int getNavyRallyMarkerCount() {
-        return navalRallyMarkers.size();
+        return NavyRallyMarkers.size();
```
Sheikah45: lower

src/java/neroxis/map/SCMap.java
```diff
@@ -341,15 +346,15 @@ public void addNavalAreaMarker(AIMarker aiMarker) {
...
     }
 
     public AIMarker getNavyRallyMarker(int i) {
-        return navalRallyMarkers.get(i);
+        return NavyRallyMarkers.get(i);
```
Sheikah45: lower

src/java/neroxis/map/SCMap.java
```diff
@@ -341,15 +346,15 @@ public void addNavalAreaMarker(AIMarker aiMarker) {
...
     }
 
     public void addNavyRallyMarker(AIMarker aiMarker) {
-        navalRallyMarkers.add(aiMarker);
+        NavyRallyMarkers.add(aiMarker);
```
Sheikah45: lower

src/java/neroxis/map/SCMap.java
```diff
@@ -376,6 +381,138 @@ public void addCubeMap(CubeMap cubeMap) {
...
+        layer5.addWithOffset(oldLayer5, locToPutCenterOfCurrentMapContent, true);
+        FloatMask layer6 =  new FloatMask(newMapBoundsSize, new Random().nextLong(), symmetrySettings);
+        layer6.addWithOffset(oldLayer6, locToPutCenterOfCurrentMapContent, true);
+        FloatMask layer7 =  new FloatMask(newMapBoundsSize, new Random().nextLong(), symmetrySettings);
+        layer7.addWithOffset(oldLayer7, locToPutCenterOfCurrentMapContent, true);
+        FloatMask layer8 =  new FloatMask(newMapBoundsSize, new Random().nextLong(), symmetrySettings);
+        layer8.addWithOffset(oldLayer8, locToPutCenterOfCurrentMapContent, true);
+
+        setTextureMasksLowScaled(layer1, layer2, layer3, layer4);
+        setTextureMasksHighScaled(layer5, layer6, layer7, layer8);
```
Sheikah45: This still seems wrong to me. seems like if the image method doesn't work for textureMasks it wouldn't work for the other images either. Would be better if the issue could be resolved and it all be done with the images.

Sheikah45: Same with the heightMap I just didn't drag far enough

src/java/neroxis/map/SCMap.java
```diff
@@ -376,6 +381,138 @@ public void addCubeMap(CubeMap cubeMap) {
...
+        for (PositionedObject positionedObject : arrayListOfPositionedObjects) {
+            float x = (float) (StrictMath.round((distanceScaler * positionedObject.getPosition().x + shiftXAndZ.x) + 0.5) - 0.5);
+            float z = (float) (StrictMath.round((distanceScaler * positionedObject.getPosition().z + shiftXAndZ.y) + 0.5) - 0.5);
+            float y;
+            if (heightmapBase.inBounds((int) x, (int) z)) {
+                y = heightmapBase.getValueAt((int) x, (int) z);
+            } else {
+                y = 0;
+            }
+            positionedObject.setPosition(new Vector3f(x, y, z));
```
Sheikah45: It should be possible to remove the dependency of heightmapBase in this function. 
The Placement class has a function placeOnHeightmap which will do lines 507-510. Additionally Vector2f has an add and subtract method. Would just need to add a multiply method and then you wouldn't need to break out the x and z components. 
Vector2f and Vector3f also have round to nearestHalfPoint method

src/java/neroxis/map/SCMap.java
```diff
@@ -376,6 +381,138 @@ public void addCubeMap(CubeMap cubeMap) {
...
+        for (int i = 0; i < getArmyCount(); i++) {
+            Army army = getArmy(i);
+            for (int a = 0; a < army.getGroupCount(); a++) {
+                Group group = army.getGroup(a);
+                repositionEachItemInAnArrayList(heightmapBase, group.getUnits(), contentScaler, shiftXAndZ);
+            }
+        }
+    }
+
+    public void repositionEachItemInAnArrayList(FloatMask heightmapBase, ArrayList<? extends PositionedObject> arrayListOfPositionedObjects, float distanceScaler, Vector2f shiftXAndZ) {
```
Sheikah45: arrayListOf is not needed in the name as it is communicated by the type.

Sheikah45: function name can just be repositionItems

Sheikah45 (changes requested):

src/java/neroxis/map/SCMap.java
Penguin5: The texture images seem to be stored differently with variable sizes.  I have literally resized around 100 times now across a large variety of maps, and the resizer works well.  Images like the heightMap appear to work just fine this way.

Sheikah45 (dismissed):

Sheikah45 (approved):

PR #265, opened 2021-01-26
New Brush-Related Content & Bulk Map Generation

Penguin5:
Added new brushes
Added Recursive Brush Generation and an option for it
Made brush generation make brushes that fillout the 512x512 more by resizing the non-empty content square to 512x512 and using that as the new brush
Added new BinaryMask and FloatMask methods/functionality to have many brush functions support input of FloatMasks if inputted
Created new FloatMask methods: trimSizeToContentAboveThreshold and resizeContentToFillVoidBelowThreshold
Re-added FloatMask methods: shrink2, enlarge2, and setSize2
Added brush selection map generation functionality
Added functionality/option to generate the desired number of maps in bulk
Added functionality to bulk generate a map for each brush (where each brush is the only brush type used on its map)


Sheikah45:
The brushes currently don't work consistently produce broken terrain

PR #279, opened 2021-11-24
Fixed guidedWalkWithBrush

Penguin5:


src/main/java/com/faforever/neroxis/mask/BooleanMask.java
```diff
@@ -369,11 +369,24 @@ public BooleanMask guidedWalkWithBrush(Vector2 start, Vector2 target, String bru
...
+            float targetY = target.getY();
+            if (wrapEdges) {
+                int maskSize = getSize();
+                int halfSize = maskSize / 2;
+                float startX = start.getX();
+                float startY = start.getY();
+                if (targetX - startX > halfSize) targetX -= maskSize;
+                if (startX - targetX > halfSize) targetX += maskSize;
+                if (targetY - startY > halfSize) targetY -= maskSize;
+                if (startY - targetY > halfSize) targetY += maskSize;
```
Sheikah45: I would use if else if and for the pairs. And also maybe just compute the difference once and use the absolute value and the sign.

PR #322, opened 2023-01-29
Add support for terrain types

BlackYps:
I added terrain type values only to one biome right now as a proof of concept


shared/src/main/java/com/faforever/neroxis/map/SCMap.java
```diff
@@ -653,6 +654,18 @@ public void setTextureMasksScaled(BufferedImage textureMasks, Vector4Mask mask)
         }
     }
 
+    public void setTerrainTypeScaled(BufferedImage terrainType, IntegerMask mask) {
```
Sheikah45: This isn't doing any scaling so that can probably be removed from the name

shared/src/main/java/com/faforever/neroxis/util/serial/biome/TerrainMaterials.java
```diff
@@ -15,4 +15,5 @@ public class TerrainMaterials {
     private String[] normalPaths = new String[TERRAIN_NORMAL_COUNT];
     private float[] normalScales = new float[TERRAIN_NORMAL_COUNT];
     private Integer[] previewColors = new Integer[TERRAIN_NORMAL_COUNT];
+    private Integer[] terrainTypes = new Integer[TERRAIN_TEXTURE_COUNT - 1];
```
Sheikah45: Rather than deriving it would rather have it be it's own explicit constant

Sheikah45 (changes requested):
The structure looks good to me although it might be good to add detection on if terrain types exist otherwise you have to add terrain types to all the biomes first.

generator/src/main/java/com/faforever/neroxis/generator/texture/BasicTextureGenerator.java
```diff
@@ -90,14 +109,16 @@ public void initialize(SCMap map, long seed, GeneratorParameters generatorParame
...
     }
 
     @Override
     public void setTextures() {
-        Pipeline.await(texturesLowMask, texturesHighMask);
+        Pipeline.await(texturesLowMask, texturesHighMask, terrainType);
         DebugUtil.timedRun("com.faforever.neroxis.map.generator", "generateTextures", () -> {
             map.setTextureMasksScaled(map.getTextureMasksLow(), texturesLowMask.getFinalMask());
             map.setTextureMasksScaled(map.getTextureMasksHigh(), texturesHighMask.getFinalMask());
+            map.setTerrainTypeScaled(map.getTerrainType(), terrainType);
```
BlackYps: Do I need getFinalMask() here? Not really sure what it does

Sheikah45: Ah yes it would be good practice. getFinalMask just guarantees all processing for that mask is finished

BlackYps:
If the terrain type is not set in the biome json and I query it like that `Integer[] terrainTypes = map.getBiome().getTerrainMaterials().getTerrainTypes();` do I get a null value or something else?

Sheikah45:
> If the terrain type is not set in the biome json and I query it like that `Integer[] terrainTypes = map.getBiome().getTerrainMaterials().getTerrainTypes();` do I get a null value or something else?

Because you specified it as Integer[] rather than int[] then the array will exist but all the elements of the array will be null.

Sheikah45:
So if 0 is a good default terrain type then you should change it to be int[]

BlackYps:
1 is the default terrain type

Sheikah45:
Then could either initialize the array with all ones or every biome needs a terraintypes array

BlackYps:
I did it with a null check now, let me know what you think

generator/src/main/java/com/faforever/neroxis/generator/texture/BasicTextureGenerator.java
```diff
@@ -73,6 +75,28 @@ protected void setupTexturePipeline() {
...
+            terrainType.add(terrainTypes[0])
+                       .setToValue(accentGroundTexture.setSize(mapSize).copyAsBooleanMask(.5f), terrainTypes[1])
+                       .setToValue(accentPlateauTexture.setSize(mapSize).copyAsBooleanMask(.5f), terrainTypes[2])
+                       .setToValue(slopesTexture.setSize(mapSize).copyAsBooleanMask(.5f), terrainTypes[3])
+                       .setToValue(accentSlopesTexture.setSize(mapSize).copyAsBooleanMask(.5f), terrainTypes[4])
+                       .setToValue(steepHillsTexture.setSize(mapSize).copyAsBooleanMask(.5f), terrainTypes[5])
+                       .setToValue(waterBeachTexture.setSize(mapSize).copyAsBooleanMask(.5f), terrainTypes[6])
+                       .setToValue(rockTexture.setSize(mapSize).copyAsBooleanMask(.5f), terrainTypes[7])
+                       .setToValue(accentRockTexture.setSize(mapSize).copyAsBooleanMask(.5f), terrainTypes[8]);
+        }
```
Sheikah45: Meh I don't think I am a big fan of the null check. I think the initialization or setting the values would be best.
I don't think it should be the generators job to validate the input here. Rather correct input should just be given.

Sheikah45 (changes requested):
Also are you sure 1 is the default value? Currently the generator has 0 everywhere as the terrain type. So it does work.

BlackYps:
Yes, 1 is the default as can be seen here: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/TerrainTypes.lua#L145 It's probably set to 1 because lua is 1-indexed.
It's obvious in the game too that 0 doesn't work, as units are missing the tread marks that they should leave on default terrain.

generator/src/main/java/com/faforever/neroxis/generator/texture/BasicTextureGenerator.java
BlackYps: How do I default to 1 values then if none are given? Or we could just skip that step alltogether and just add terrain types to all biomes. Could be all default values for starters

Sheikah45:
Got it so 0 doesn't crash the game it just doesn't do anything

generator/src/main/java/com/faforever/neroxis/generator/texture/BasicTextureGenerator.java
Sheikah45: You would just initalize the array in terrainMaterials as an array of all 1s. Or could just make it an int[] so that it is more obvious when a biome doesn't have any terrain types set.

Allthough I guess I have to check if the serializer we currently use would overwrite the value. If it does we would have to make a setter that checks in terrainMaterials

BlackYps:
yes

generator/src/main/java/com/faforever/neroxis/generator/texture/BasicTextureGenerator.java
BlackYps: seems like the serializer only overwrites the default values if he finds actual values in the json

Sheikah45: Then great we can just use the initialization to set default values

BlackYps: I already pushed the change. Have a look if it is implement in the way you imagined

Sheikah45 (approved):
Were you planning on adding the terrain types for the other biomes?

Because if they aren't added now they probably never will be

BlackYps:
I'm planning to either do it myself or find someone interested in doing it. I agree that if we don't do it soon it will probably never happen. The change to the working default type should alredy be a big improvement though

Sheikah45:
Guess that means I need to actually make a release ha. Although I need to fix a few other things first before that

PR #323, opened 2023-01-30
Add terrain types for all biomes

BlackYps:
I fixed the symmetry problem, but not in a good way I think. It's probaby better to have a setSize method that doesn't meddle with the symmetry, but I don't understand that method and the surrounding code enough to do that.


generator/src/main/java/com/faforever/neroxis/generator/texture/BasicTextureGenerator.java
```diff
@@ -81,17 +82,32 @@ protected void setupTexturePipeline() {
...
+        // setSize enforces symmetry, so we need to get rid of the symmetry settings first
+        SymmetrySettings noSymmetry = new SymmetrySettings(Symmetry.NONE);
+        accentGroundTexture.setSymmetrySettings(noSymmetry);
+        accentPlateauTexture.setSymmetrySettings(noSymmetry);
+        slopesTexture.setSymmetrySettings(noSymmetry);
+        accentSlopesTexture.setSymmetrySettings(noSymmetry);
+        steepHillsTexture.setSymmetrySettings(noSymmetry);
+        waterBeachTexture.setSymmetrySettings(noSymmetry);
+        accentRockTexture.setSymmetrySettings(noSymmetry);
+        rockTexture.setSymmetrySettings(noSymmetry);
```
Sheikah45: The symmetry should be set at construction

Sheikah45 (changes requested):

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
```diff
@@ -37,6 +37,7 @@
     @Getter
     private final String name;
     @Getter
+    @Setter
     protected SymmetrySettings symmetrySettings;
```
Sheikah45: The symmetrySettings should be marked final so no setter

BlackYps:
Should be good now

Sheikah45 (dismissed):

Sheikah45 (approved):

PR #324, opened 2023-02-03
Set cubemaps from biome

BlackYps:


Sheikah45 (changes requested):
CubeMaps should not be set in the texture generator. They are already included in the biome which is a member of the scmap already. So if they can be set directly in the SCMap constructor.

Although really looking at this I think it may be better to just have the cubemaps directly in the biome and serialize and deserialize them. Rather than having the name and path separate.

BlackYps:
So how would I do that? How exactly should they be stored in the biome?

BlackYps:
It sounds similar to what I did in the first commit

BlackYps:
I think I got it

shared/src/main/java/com/faforever/neroxis/map/SCMap.java
```diff
@@ -120,7 +120,9 @@ public SCMap(int size, Biome biome) {
         navalRallyMarkers = new ArrayList<>();
         waveGenerators = new ArrayList<>();
         cubeMaps = new ArrayList<>();
-        cubeMaps.add(new CubeMap("<default>", "/textures/environment/defaultenvcube.dds"));
+        for (CubeMap cubeMap : biome.getTerrainMaterials().getCubeMaps()) {
+            addCubeMap(cubeMap);
+        }
```
Sheikah45: Since the biome contains the cubeMaps we can remove the addCubeMap method and just use the ones embedded in the biome for import and export

Sheikah45 (changes requested):

shared/src/main/java/com/faforever/neroxis/util/serial/biome/TerrainMaterials.java
```diff
@@ -1,15 +1,19 @@
...
+import java.util.List;
+
 @Data
 @CompiledJson
 public class TerrainMaterials {
     // engine limitations - must stay 9 and 10 always
     public static final int TERRAIN_TEXTURE_COUNT = 10;
     public static final int TERRAIN_NORMAL_COUNT = 9;
     private String name;
+    private List<CubeMap> cubeMaps;
```
Sheikah45: This should be initialized otherwise the importer will throw a null pointer exception

Sheikah45 (changes requested):

BlackYps:
lol, the build fails now, but I don't understand why that commit could break it and it works locally on my machine...

Sheikah45 (approved):

PR #337, opened 2023-03-29
Improve water, lighting and decals

BlackYps:
Short summary:
I changed the water ramp on maps that are not a green biome.
Some biomes had almost pitch-black shadows, I made them a bit brighter,
generally made the lighting on Frithen brighter,
changed decals that looked out of place,
lowered the abyss level, so maps don't lose all shading in deeper oceans.


Sheikah45:
Please post pictures of all the changed biomes before and after

BlackYps:
Desert
Screenshot 2023-04-01 150349
Screenshot 2023-04-01 155523
Screenshot 2023-04-01 155912
Screenshot 2023-04-01 161153
Screenshot 2023-04-01 155850
Screenshot 2023-04-01 155556

Frithen
Screenshot 2023-04-01 150306
Screenshot 2023-04-01 160604
Screenshot 2023-04-01 160236
Screenshot 2023-04-01 160825

BlackYps:
Wonder
Screenshot 2023-04-01 150726
Screenshot 2023-04-01 162110

Stones
Screenshot 2023-04-01 150644
Screenshot 2023-04-01 162213

Brimstone
Screenshot 2023-04-01 151243
Screenshot 2023-04-01 162726

Syrtis
Screenshot 2023-04-01 150759
Screenshot 2023-04-01 161902

Loki
Screenshot 2023-04-01 151023
Screenshot 2023-04-01 162540

BlackYps:
Prayer
Screenshot 2023-04-01 150850
Screenshot 2023-04-01 161702

Winding River
Screenshot 2023-04-01 150431
Screenshot 2023-04-01 162422

Early Autumn
Screenshot 2023-04-01 150601
Screenshot 2023-04-01 162308

BlackYps:
Moonlight
Screenshot 2023-04-01 150524
Screenshot 2023-04-01 214144
Screenshot 2023-04-01 213832
Screenshot 2023-04-01 214028

Mars
Screenshot 2023-04-01 150939
Screenshot 2023-04-01 212201
Screenshot 2023-04-01 213229
Screenshot 2023-04-01 212937

The water in every biome now has foam like on Mars, I just didn't make screenshots of all of them as it seemed redundant. The waves are much slower now as well.

BlackYps:
I have now adjusted everything that I wanted to adjust. From my POV it's ready for review.

Sheikah45:
I gotta say the ones where the water turned brown I am not really fond of. To me it just becomes hard to see the water and they look more like holes in the ground.

Specifically: Desert, Mars, and maybe Stones, but I could go either way on that.

Sheikah45:
Also for Frithen I will have to check but it might be too bright now. A reason it was made more dim was because of complaints about it being blinding

BlackYps:
I tried to create a consistent visual impression in the biomes. I think the blue water on everything looks super lazy especially as it is the default water ramp that is overused as hell. I don't think water has to be strictly blue to be recognisable as water. That said I am open for feedback, but then we should probably gather some more opinions than just us two.

BlackYps:
> Also for Frithen I will have to check but it might be too bright now. A reason it was made more dim was because of complaints about it being blinding

I didn't have the impression that it is too bright. When playing the fog of war makes it even darker for the most part, but I generally don't have too much of a problem with bright maps, so maybe I am the wrong person to ask

Sheikah45:
I think for Mars and Desert it is just that it doesn't seem to fit any more because the water looks very clear but then the ocean floor is such a different color from the surface. It is a little jarring. Maybe if the water looked murkier or the floor was just a bit brighter?

BlackYps:
Yeah, I was trying to go for murky water, but it's hard to pull off with the options we got. Do you have an idea how to pull it off better? Or examples of other maps that do it?

Sheikah45:
I don't know of any examples because this is the first attempt I have seen. I imagine that it could be done by making the shallow depth be much less prominent or increasing the lerp.

BlackYps:
I had a look at the gpg maps and I noticed that they use a different texture for everything that is below the water surface. That helps sell the murky water look, because they use a darker texture. Unfortunately the map generator currently uses one texture to paint the beaches and ocean floor, so we would have to change that first.

Garanas:
The decal changes are great, especially on `Moonlight` the old decals quickly look out of place because they flatten the normals of the terrain. In general I very much look forward to these changes! They add a great deal of diversity to the biomes that map authors usually apply to their maps too. To dive into the feedback of Sheikah:

`High Noon` is a desert-like map from the original game:

image

This is what it looks like without water:

image

You can 'cheat' this behavior by creating a mask that is only applied 20% (which would be 60% opacity due to shader bug where stratum layers start at 50%), and then applied more and more as the water gets deeper. You can create the 'muddy' feeling through that. Increase the abyss depth (so that the gradient is more gradually applied) and then toy with half-applied stratum layers (20%, 40%, 60%). You can use other textures for this as either it is so barely visible that you just use it to tune the terrain color or it is so much underwater that the water ramp takes over. You can try this with the rock texture maybe, I'm not sure which textures are available for the desert biome.

For Mars I'd slightly increase the surface color / decrease the surface transparency. It is difficult to see the water line.

phongu:
The new colors look so much better!

BlackYps:
Can you elaborate on the shader bug? I think the map generator doesn't take it into account

Garanas:
Only the last 50% of the stratum mask (127 up to 255) are used. Everything below 127 counts as 0

BlackYps:
Looks like we should fix that then, but that should be a separate PR in my opinion

Garanas:
I think the map generator already takes it into account - @Sheikah45 could you shed some light on this?

Sheikah45:
The fact that the texture intensity only goes from 128-255 yes we account for that.

BlackYps:
Yeah, it's not really explicitly documented, but I noticed that some of the texture layers add .5  or something similar when generating the masks

BlackYps:
I recycled the texture slot for the steepHills as it contributed almost nothing visually into a texture for the underwater areas. For most biomes I just copied the beach texture into them as there are no problems discerning the water line. So visually they are basically unchanged. For the rest I made some screenshots. I think this should alleviate the problems with the water not looking murky enough or the shore line being hard to see.

Screenshot 2023-04-21 122653
Screenshot 2023-04-21 122808
Screenshot 2023-04-21 124510
Screenshot 2023-04-21 123142
Screenshot 2023-04-21 124112

BlackYps:
With these last changes to Wonder I think everything is set now. I plan to rename some biomes, but this should better be done in a separate pull request as it makes the changes in the files much harder to figure out.

Screenshot 2023-04-21 202040

Sheikah45 (approved):

PR #343, opened 2023-04-21
Rename biomes

BlackYps:
Depends on #337


Sheikah45:
The only issue with changing biome names is that someone may be externally using them in the command line so now that will fail. What is your thought process with changing them?

It looks like you want to just make them more generic.

BlackYps:
In general I want to move away from using map names as biome names. It requires the user to know the original map to have any idea what to expect from a biome. On top of that some of the biomes are only roughly based on the map they refer to, so it makes even less sense.
I'll give my reasonings for the individual renames:
I can't for the life of me remember which one is Prayer and which one is Winding River. I have no connection to either of these maps and it is extremely weird to me to name a biome after a topological feature that isn't even present in the generated maps. I chose Evergreen as the new name as that is the name of the texture folder that it utilizes. Green pastures was just an association by me based on the looks as Evergreen was already taken.
The Frithen biome deviates quite a bit from the map Frithen and I wanted to create an opportunity to add a snowy biome as this biome uses very "icey" textures, so I named it Ice.
Syrtis also deviates from the map Syrtis enough to not form an immediate connection in my head, so I named it like the texture folder that it uses.
I transformed Wonder into a swamp biome, so might just call it like that. And it definitely doesn't have the "Wonder" look anymore.
Moonlight is a bit tricky as it has a very distinct look and it's hard to name what the material could be, but I wanted to move away from the map names, so it turned into Toxic Sludge.
I would be ok with keeping Loki as a name, as it is almost a 1 on 1 copy and the map is very known, especially as Sunset is not the best name either in my opinion, so I would be happy about suggestions.

BlackYps:
I don't really think that someone has set up an elaborate external script that we would be breaking, so I am willing to take that risk. On top of that we have also changed the looks of some of the biomes considerably, so while not the most elegant way to inform people about that change it also doesn't sound like the worst occasion.

Sheikah45:
Maybe but I am also just thinking that the generic names are rather limited. For example of someone adds another ice/snow biome in the future they will have to name it something exotic anyway to differentiate it.

So the name space to me needs to be sparse to allow for that. Additionally the using map names was just to indicate what inspired it.

So I am not entirely convinced yet that things need to move to more generic names.

BlackYps:
I think there are enough synonyms left that we use to describe landscapes. For the icey stuff tundra, winter, permafrost immediatly come to my mind.
I agree that it's not terribly important to rename them, but I think it is good user experience to give them names that are actually helpful to understand or memorize what the biome looks like. Otherwise we could just give them numbers. The usage of map names tried to do that, but I already explained what the problems are with that approach.

Sheikah45:
At the moment the biome names are not really too exposed to general users in any fashion. So I am not really worried about user experience with people knowing what the name means. 

For now I think it is nicer to have the little easter eggs of names with almost the idea that the biomes are named after the planets they come from. Allows a little creativity as well.

BlackYps:
I guess for now we can leave it as is, but in the future it would be nice to be able to select the biome in the client ui and then we need good names

PR #348, opened 2023-04-29
Tweaked Biomes

Penguin5:
Adjusted biomes: Desert, Frithen, Loki, Mars, Moonlight, Prayer, Stones, and Syrtis
Changes were primarily to water settings and lighting

### DIFF
--- a/shared/src/main/resources/custom_biome/Desert/Light.scmlighting
+++ b/shared/src/main/resources/custom_biome/Desert/Light.scmlighting
@@ -1,37 +1,37 @@
 {
-  "lightingMultiplier" : 0.86,
-  "bloom" : 0.08,
-  "fogStart" : 0.0,
-  "sunDirection" : {
-    "z" : 0.1263912,
-    "x" : 0.7083398,
-    "y" : 0.7058716
+  "sunColor": {
+    "z": 1.5,
+    "y": 1.45,
+    "x": 1.39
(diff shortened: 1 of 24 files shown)


Sheikah45:
Need pictures of before and after. Although if the changes are purely personal opinion there is no guarantee on them getting merged as otherwise this just turns into a back and forth of prs which I do not want.

Penguin5:
Images posted in order of old then new:
Mars
image
image
Frithen
image
image
Loki
image
image
Desert
image
image
Moonlight
image
image
Syrtis
image
image
Prayer
image
image
Stones
image
image

Sheikah45:
This looks like it is just reverting the color changes that blackyps made. I think this only reduces the variance of the biomes so will not be merged.

Penguin5:
I spent time manually adjusting each of those.  I didn't revert it to an old version.  I'd like the changes to actually be considered rather than just unilaterally swept aside.  I can understand not merging some of the changes and or suggesting further adjustments to them, but I think unilaterally rejecting all of them is the wrong way to handle this situation.

Sheikah45:
All this appears to have done is set the colors to all be the same blue. The purpose of changing the water was to give it more variety. This pr removes the variety and that seems to be your intent. So there are no comments beyond it is not in the spirit of where I think these biomes should be.

PR #351, opened 2023-05-09
Export pbr custom texture

Sheikah45:


BlackYps:
This should be fine for now

Sheikah45:
The build is failing right now

BlackYps:
the old problem of not having pulled all commits xD

BlackYps:
the changed lightsettings are necessary to inform the unit shader that we use exponential water absorption. See https://github.com/FAForever/fa/pull/5172

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
```diff
@@ -523,6 +525,23 @@ private void generate() {
...
+            map.getBiome().terrainMaterials().getNormalPaths()[8] =
+                    map.getBiome().terrainMaterials().getCubeMaps().get(0).getPath();
+        }
+        map.getBiome().terrainMaterials().getTexturePaths()[9] =
+                Path.of("/maps", map.getFolderName(), "env", "texture", "mapwide.dds")
+                                       .toString()
+                                       .replace("\\", "/");
+        // This needs to be higher than the map size in ogrids to trigger all aspects of the terrain
+        // shader, so we set this to a sufficiently high value.
+        map.getBiome().terrainMaterials().getTextureScales()[9] = 10000;
```
Sheikah45: What does this mean? So is this not based on map size at all?

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
```diff
@@ -523,6 +525,23 @@ private void generate() {
...
+                    Path.of("/maps", map.getFolderName(), "env", "texture", "heightRoughness.dds")
+                            .toString()
+                            .replace("\\", "/");
+            map.getBiome().terrainMaterials().getNormalPaths()[8] =
+                    map.getBiome().terrainMaterials().getCubeMaps().get(0).getPath();
+        }
+        map.getBiome().terrainMaterials().getTexturePaths()[9] =
+                Path.of("/maps", map.getFolderName(), "env", "texture", "mapwide.dds")
+                                       .toString()
+                                       .replace("\\", "/");
```
Sheikah45: Why is only the heightRoughness dependent on the PBR Shader?

shared/src/main/java/com/faforever/neroxis/exporter/SCMapExporter.java
```diff
@@ -231,44 +229,29 @@ public static void exportPreview(Path folderPath, SCMap map) throws IOException
...
+        Path writingPath = folderPath.resolve(filePath);
+        Path sourcePath = Paths.get("shared/src/main/resources/images/heightRoughness.dds");
         Files.createDirectories(writingPath.getParent());
-        Decal shadowDecal = new Decal(decalParent.resolve(decalPath).toString().replace('\\', '/'), topLeftOffset,
-                                      new Vector3(), size, 1000);
-        shadowDecal.setType(DecalType.WATER_ALBEDO);
-        map.getDecals().add(shadowDecal);
         try {
-            Files.write(writingPath, compressedShadows, StandardOpenOption.CREATE);
+            Files.copy(sourcePath, writingPath, StandardCopyOption.REPLACE_EXISTING);
```
Sheikah45: Are you trying to copy from the repo code? If so this won't work when it is packaged as a jar. You have to get the classpath resource like we do with biomes.

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -228,10 +228,24 @@ Vector3 calculateNormalAt(int x, int y, float scale) {
...
+     */
+    @GraphMethod
+    public FloatMask normalize() {
+        return enqueue(() -> {
+            float min = getMin();
+            float max = getMax();
+            subtract(min);
+            divide(max - min);
+        });
+    }
```
Sheikah45: If we make this a dedicated method then just use apply and use the primitive functions to do the subtraction and division in one step and avoid autoboxing

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
BlackYps: what do you mean?
81km maps are 4096x4096, so the value needs to be higher. 5000 would also suffice, but it doesn't hurt when it is 10,000.

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
BlackYps: because we use the mapwide assets for the Terrain002 shader as well, not just for the pbr shader

shared/src/main/java/com/faforever/neroxis/exporter/SCMapExporter.java
BlackYps: so I would put the image file into the biome folder, or do you mean something else?

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
BlackYps: oh actually this is leftover code and can be removed

shared/src/main/java/com/faforever/neroxis/exporter/SCMapExporter.java
Sheikah45: No you need to load the file using the same mechanism we load the biome files with.
That works with a directory or a jar

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
Sheikah45: So shouldn't there be a check for Terrain002 shader?

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
Sheikah45: It would make more sense if it just needs to be bigger or the same size as the map to just set it to the mapsize or mapsize+1 rather than an arbitrary value

BlackYps:
I noticed a problem, the generated normal map is unnecessarily blurred.
original:
Screenshot 2023-09-04 210435

ours:
Screenshot 2023-09-04 210421

The small hill at the bottom is indeed a bit rough, which is completely blurred with our generated normal map. But you can see it around the plateau as well

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
BlackYps: oh yeah

shared/src/main/java/com/faforever/neroxis/exporter/SCMapExporter.java
BlackYps: ImageIO can't deal with dds files. What should I do?

BlackYps: Oh wait, I think I could provide the source file as png and just write it as dds with the Imageutil class

shared/src/main/java/com/faforever/neroxis/exporter/SCMapExporter.java
```diff
@@ -231,44 +227,34 @@ public static void exportPreview(Path folderPath, SCMap map) throws IOException
...
+            BufferedImage image = ImageIO.read(sourceURL);
+            Path textureDirectory = Paths.get("env", "texture");
+            Path filePath = textureDirectory.resolve(PBR_DDS);
+            Path writingPath = folderPath.resolve(filePath);
+            Files.createDirectories(writingPath.getParent());
+            try {
+                ImageUtil.writeNormalDDS(image, writingPath);
+            } catch (IOException e) {
+                System.out.print("Could not write the pbr texture\n" + e);
+            }
```
Sheikah45, line 255: Why are we reading it in as an image of we are just copying it? You can just get it as a resourcestream and write that input stream using the nio files API.

Sheikah45:
The normal map is smoothed because otherwise artifacts appear

BlackYps:
what kind of artifacts?

shared/src/main/java/com/faforever/neroxis/exporter/SCMapExporter.java
BlackYps: because I wasn't aware of that solution :D

but I got some post-hoc justifications:
this way it would be easier later to edit the file, as the source file is a png and the file that the map reads should be dds

Sheikah45: Is it just a raw DDS file with no compression or mip mapping?

Sheikah45:
It just becomes too sharp with edges

shared/src/main/java/com/faforever/neroxis/exporter/SCMapExporter.java
BlackYps: I think both don't really matter. Compression shouldn't be an issue and the game would generate the mipmaps if they aren't in the file already. I guess having the mipmaps already would improve loading times, but I don't know if the time for that is more than even a fraction of a second

Sheikah45:
@BlackYps are the shader names correct on this. If so then this should be ready

BlackYps:
Yes, they are correct

BlackYps:
I noticed that I made an error when converting the light settings. They are fixed now, so while the values have changed, the visuals stay the same

BlackYps (approved):

PR #360, opened 2023-08-26
Biome water update

BlackYps:
This adjusts the water by using better wave textures and making the sun actually reflect in the water. This might make it necessary to adjust the position of the sun, to produce nicer sun reflections for default camera angles, but this needs more testing to determine.


BlackYps:
Looking good now

BlackYps:
Superseded by #371

PR #365, opened 2023-10-08
Pbr texture generator

BlackYps:
This PR adds three things:
1) We write the water depth in the previously unused channel of the mapwide texture.
At the moment this has no effect, but I plan to do a map shader that uses this. The shader that the generator now uses doesn't utilize that channel, so writing something in there has no side effects.
2) We add a tool to generate this texture for a provided map.
This makes it easier for mapmakers to upgrade their maps from the default shader to Terrain002, that requires this texture.
3) We add a tool to generate the pbr map from individual height and roughness maps.
Again this helps mapmakers with setting up advanced shaders.

Can these tools be accessed when you only have a compiled generator or do people need to download the repo to use them?


shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
```diff
@@ -135,14 +135,121 @@ public static void writeNormalDDS(BufferedImage image, Path path) throws IOExcep
         ddsHeader.setFourCC("DXT5");
         byte[] compressedData = compressImage(imageBytes.array(), ddsHeader.getWidth(), ddsHeader.getHeight(), null,
                                               Squish.CompressionType.DXT5);
+        if (Files.exists(path)) {
+            // If we don't do this we get weird results when the file already exists
+            Files.delete(path);
+        }
```
Sheikah45: Files.deleteIfExists

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
```diff
@@ -135,14 +135,121 @@ public static void writeNormalDDS(BufferedImage image, Path path) throws IOExcep
...
+        if (baseImage.getWidth() != heightMap.getWidth() * 4) {
+            throw new IllegalArgumentException((
+                    "Texture sizes are not compatible: base image width %d, height map width %d. " +
+                    "The height map needs to be quarter the resolution of the base image."
+            ).formatted(baseImage.getWidth(), heightMap.getWidth()));
+        }
+        int channel = 0;
+        if (layer >= 4) {
+            channel = 2;
+        }
```
Sheikah45: What is this logic?

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
```diff
@@ -135,14 +135,121 @@ public static void writeNormalDDS(BufferedImage image, Path path) throws IOExcep
...
+        if (baseImage.getWidth() != roughnessMap.getWidth() * 4) {
+            throw new IllegalArgumentException((
+                    "Texture sizes are not compatible: base image width %d, roughness map width %d. " +
+                            "The roughness map needs to be quarter the resolution of the base image."
+            ).formatted(baseImage.getWidth(), roughnessMap.getWidth()));
+        }
+        int channel = 1;
+        if (layer >= 4) {
+            channel = 3;
+        }
```
Sheikah45: What is this logic?

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/MapEnvTextureExporter.java
```diff
@@ -0,0 +1,69 @@
...
+        return 0;
+    }
+    
+    public void generateEnvTexture() throws Exception {
+        System.out.print("Generating env texture\n");
+        SCMap map = MapImporter.importMap(requiredMapPathMixin.getMapPath());
+
+        // Copied from TextureGenerator and BasicTextureGenerator
+        FloatMask heightMapSize = new FloatMask(map.getHeightmap(), (long) 0, new SymmetrySettings(Symmetry.NONE))
+                .startVisualDebugger()
```
Sheikah45: startVisualDebugger should be removed

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/MapEnvTextureExporter.java
```diff
@@ -0,0 +1,69 @@
...
+    public Integer call() throws Exception {
+        generateEnvTexture();
+        return 0;
+    }
+    
+    public void generateEnvTexture() throws Exception {
+        System.out.print("Generating env texture\n");
+        SCMap map = MapImporter.importMap(requiredMapPathMixin.getMapPath());
+
+        // Copied from TextureGenerator and BasicTextureGenerator
```
Sheikah45: This is an unnecessary comment

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/MapEnvTextureExporter.java
```diff
@@ -0,0 +1,69 @@
...
+        FloatMask heightMapSize = new FloatMask(map.getHeightmap(), (long) 0, new SymmetrySettings(Symmetry.NONE))
+                .startVisualDebugger()
+                .resample(map.getSize())
+                .divide(128f); // No idea why this is necessary
```
Sheikah45: I am not sure why you are doing this either. Normally it has to do with the scaling of the mask values but I am not sure what they are supposed to be here

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/MapEnvTextureExporter.java
```diff
@@ -0,0 +1,69 @@
...
+        NormalMask normals = heightMapSize.copy()
+                .copyAsNormalMask(2f);
```
Sheikah45: don't need a copy of a copy

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
```diff
@@ -0,0 +1,61 @@
...
+                        if (path.getFileName().toString().startsWith("Roughness")) {
+                            System.out.print("Writing roughness texture\n");
+                            pbrTexture = ImageUtil.setRoughnessMap(pbrTexture, image, layer);
+                        } else if (path.getFileName().toString().startsWith("Height")) {
+                            System.out.print("Writing height texture\n");
+                            pbrTexture = ImageUtil.setHeightMap(pbrTexture, image, layer);
+                        }
+                    }
+                }
+            }
```
Sheikah45: Rather than searching for the files with pattern matching we should just have the user pass the files via a commandline option

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
```diff
@@ -135,14 +135,121 @@ public static void writeNormalDDS(BufferedImage image, Path path) throws IOExcep
...
+                            value /= 256;
+                        }
+                        int[] baseImagePixel = baseImageRaster.getPixel(newX, newY, new int[4]);
+                        baseImagePixel[channel] = value;
+                        baseImageRaster.setPixel(newX, newY, baseImagePixel);
+                    }
+                }
+            }
+        }
+        return baseImage;
```
Sheikah45: Is this just writing masks into images with some offset? Util methods already exist for this kind of operation. So I guess need a little more insight onto why this needs it's own special method. 

It is also named as though it is general when there is some very special operations going on here.

Sheikah45 (changes requested):
A separate Jar file needs to be built for users to use it. The build is there it just has never been deployed via github

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
BlackYps: Isn't this quite inconvenient when you have to pass up to 16 files? It creates a rather long input string which is a bit clunky to handle
What is the downside of doing it this way?

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
BlackYps: Yes, it is writing into the image with an offset that I calculate based on the layer number. I need to control which channel we are writing into because I need to write into one channel only per mask. The `insertImageIntoNewImageOfSize` method looked like it would write a greyscale image into all the channels of the bigger image.

Good catch that it should not have such a generic name.

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/MapEnvTextureExporter.java
BlackYps: The mask values need to be the actual height of the heightmap, so around 0 to 50. This is needed so the normal map has the right strenght and the calculations for the water depth produce the correct results.
For some reason the heightmap values are initally to big by a factor of exactly 128.

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
BlackYps: It's for controlling which channel I write the mask into. See also my other comment below for more explanation.

BlackYps:
Here is an example of a resulting pbr texture, so you can get a better understanding of what is happening
heightRoughness

Note that a given input texture is in the middle of a quadrant like so:
grafik

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/MapEnvTextureExporter.java
Sheikah45: The 128 is because in the scmap binary the scale factor is 128. And if it is just about the normal map you could use the scalar argument to copy as normal mask

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
Sheikah45: I would view it a bit more favorably than right now which is an opaque naming scheme one would have to follow with no failure mode in case one is not found

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
Sheikah45: It isn't only grayscale, or shouldn't be it just depends on the image types themselves. And it can can larger or smaller or offset.

Also you could instead just create a vector 4 mask as well and then save that as an image using normal utilities.

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
Sheikah45: At least we should have the user specify the folder or name prefix and throw errors if any assumptions we make about which files are present ends up being false.

Otherwise users just get not the result they want and it is hard to figure out why exactly.

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/MapEnvTextureExporter.java
BlackYps: It's also relevant for the water depth

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
BlackYps: At the moment the user specifies the folder where to look for the textures. Should I create an informative log message when I skip a file because it doesn't fit the naming scheme? I agree that we should try to avoid that it is too opaque, but at the moment I am not sure what kind of solution you have in mind. Or where we should inform the user

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
BlackYps: Do you think doing these operations on a mask would be simpler?

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
Sheikah45: I would check if all the expected files actually exist and ideally that there are no other files in the directory and if so throw and I formative exception

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
Sheikah45: Yes because then you could work easily on individual layers instead of all the conditional checks about which layer are we actually working on.

Sheikah45: And the offset methods already exist rather than having another version.

Ideally all processing is done on masks and then persistence is done in images. It keeps the logic and abstraction layers simpler.

BlackYps: doing it directly with images seemed easy because I already load images.
Which methods would I use to set a single component of the Vector4Mask? And how can I copy a smaller mask into a bigger mask with offset? I couldn't find any suitable methods.

Sheikah45: There are the withOffset methods and Component methods that can be used.

BlackYps: I'm afraid I can't really wrap my head around the stuff that goes on in the masks. I started using them for the processing, but I get index out of bounds errors. 2048 is obviously out of bounds when the array is 2048 elements long, but I can't figure out why the index grows to that, the loop method seems correct. You can have a look by running the pbr texture generator.
I also have to write a function that writes the vector mask into an image, but that is beyond my abilities.

shared/src/main/java/com/faforever/neroxis/mask/VectorMask.java
```diff
@@ -522,6 +522,14 @@ public U addComponent(float value, int component) {
         return addComponent((x, y) -> value, component);
     }
 
+    @GraphMethod
+    public U setComponent(FloatMask other, int component) {
+        return enqueue(dependencies -> {
+            FloatMask source = (FloatMask) dependencies.get(0);
+            setComponent(source::get, component);
```
Sheikah45: this could be getPrimitive

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
Sheikah45: Can we change this method name

Sheikah45: Also this can just be a ternary operator assignment

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
Sheikah45: ternary operator

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
Sheikah45: I can take a look in maybe a few days. Also there is a method that writes a vector mask to an image already. Does that not work?

Sheikah45 (changes requested):

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
BlackYps: No rush, it's not urgent.
That method does some processing on the values, which are probably couterproductive for what I want to do, but I would have to look closer.
I have been thinking again about the approach to do the processing in masks. I can see the use when you start out with masks, then it makes sense to keep all the processing in the masks and only to convert on the last step.
But in this case we start out with images, and we don't do any processing on the values, just writing the values into a different area. When looking into the masks I noticed that I would have to write new methods for some of the steps I need to do anyway and it's 10 times harder for me because the masks are doing a lot of pretty abstract stuff that I need to understand first before I can make a call what methods to use or how to write my new methods.
On the contrary the approach with the image manipulation works already. I wonder if it's really worth it to basically reimplement the whole thing to satisfy the abstract notion of "processing should be done in masks". In my personal opinion it isn't, partly because that rewrite would take significantly longer than fixing my first implementation would take.

I concede that my first implementation is not the most readable, but I think this can be refactored relatively easily. Maybe it is also a good idea to not store the methods in imageUtil, but in the tool directly, as they are pretty special in what they do and not really suitable for other use cases.

Sheikah45: My main thinking is that I don't want the processing aspects to be split into multiple places. ImageUtil is really for basic operations centered around converting masks to images.

I more so want to stay away from having duplicate operations one for masks and one for images as that can be a maintenance pain and also with images there are a lot of caveats that one would have to make assumptions about which makes it less applicable and easier to fall into traps.

So that is why I prefer processing is done in masks because there are less assumptions about what the is required and the then can be easily translated.

And what steps would you need new methods for?

BlackYps: I thought I needed to write a new method for writing the vector4Mask to an image without processing, because toImage() does a lot of processing, but I now found writeToImage() from the vectorMask class.
I guess it would also be nicer to use setWithOffset instead of addWithOffset, because that is closer to what is supposed to happen, but it doesn't matter when performing this operation on a floatMask that is 0 everywhere.

BlackYps: Can you try to run the PbrTextureGenerator and help me with that out of bounds error? I don't know why this happens. If that error is overcome then I think I can do the rest.

Sheikah45: We can add setWithOffset it should be there. I probably just got lazy

Sheikah45: And yeah sorry guess I should have been more specific that method already existed

shared/src/main/java/com/faforever/neroxis/mask/VectorMask.java
```diff
@@ -668,4 +677,84 @@ public FloatMask[] splitComponentMasks() {
...
+    }
+
+    public U divideComponentWithOffset(FloatMask other, int component, int xOffset, int yOffset, boolean center, boolean wrapEdges) {
+        return enqueue(dependencies -> {
+            FloatMask source = (FloatMask) dependencies.get(0);
+            applyWithOffset(source, (BiIntFloatIntConsumer) this::divideComponentAt, component, xOffset, yOffset, center, wrapEdges);
+        }, other);
+    }
+
+    private U applyWithOffset(FloatMask other, BiIntFloatIntConsumer action, int component, int xOffset, int yOffset, boolean center, boolean wrapEdges) {
```
Sheikah45:
```suggestion
    private U applyComponentWithOffset(FloatMask other, BiIntFloatIntConsumer action, int component, int xOffset, int yOffset, boolean center, boolean wrapEdges) {
```

Sheikah45 (approved):

PR #371, opened 2023-11-05
Use Pbr shader

BlackYps:
Uses new biomes, because the texture order is drastically different


shared/src/main/java/com/faforever/neroxis/map/SCMap.java
```diff
@@ -663,7 +660,17 @@ public void setTerrainType(BufferedImage terrainType, IntegerMask mask) {
                 terrainType.getRaster().setPixel(x, y, new int[]{val0});
             }
         }
+    }
 
+    public void setWaterSurfaceShadow(BufferedImage waterShadowMap, FloatMask mask) {
```
Sheikah45: I would make the name of this setter consistent with the name of the property. So either should be setWaterShadowMap or the parameter should be waterSurfaceShadow

toolsuite/src/main/java/com/faforever/neroxis/toolsuite/MapForcer.java
```diff
@@ -127,7 +127,7 @@ private void forceTerrain(SCMap map) {
         IntegerMask normalMask = new IntegerMask(map.getNormalMap(), null, symmetrySettings, "normal");
         IntegerMask waterMask = new IntegerMask(map.getWaterMap(), null, symmetrySettings, "water");
         IntegerMask waterFoamMask = new IntegerMask(map.getWaterFoamMap(), null, symmetrySettings, "waterFoam");
-        IntegerMask waterFlatnessMask = new IntegerMask(map.getWaterFlatnessMap(), null, symmetrySettings,
+        IntegerMask waterFlatnessMask = new IntegerMask(map.getWaterShadowMap(), null, symmetrySettings,
```
Sheikah45: This name should change too

shared/src/main/java/com/faforever/neroxis/biomes/Biomes.java
```diff
@@ -21,7 +21,7 @@ public class Biomes {
     // ├-- WaterSettings.scmwtr <required>
     // └-- Light.scmlighting <required>
     public static final List<String> BIOMES_LIST = List.of("Brimstone", "Desert", "EarlyAutumn", "Frithen",
-                                                           "Loki", "Mars", "Moonlight", "Prayer", "Stones",
+                                                           "PBR_Loki", "Mars", "Moonlight", "Prayer", "Stones",
```
Sheikah45: If we are wholesale replacing loki we don't need the PBR prefix

shared/src/main/java/com/faforever/neroxis/util/serial/biome/TerrainMaterials.java
```diff
@@ -33,16 +21,5 @@ public class TerrainMaterials {
...
-            1,  // plateauAccent
-            1,  // slopes
-            1,  // slopesAccent
-            40,  // beach
-            150,  // rock
-            150,  // rockAccent
-            221,  // shallowWater
-            223   // deepWater
-    };
+    private Integer[] terrainTypes = {};
```
Sheikah45, line 24: We should document what the expected number of terrain types is here in the initialization

Sheikah45 (changes requested):
The height roughness got larger XD

shared/src/main/java/com/faforever/neroxis/biomes/Biomes.java
BlackYps: whoops, that was not supposed to happen

Sheikah45: I would say we could just wholesale replace it rather than have PBR prefix

BlackYps: Then I think we should rename it to something else entirely. It's inspired by Loki, but at this points it's quite different.

BlackYps:
We should probably replace the png with a dds. I noticed that the time it takes to compress the dds file is even longer than my machine takes for the whole map generation of a 20km map.
The only open question I see is, how do we keep it editable for the future. Every time you modify a dds and resave it, the compression degrades it further. So we need an original png file somewhere. That doesn't have to be shipped in the jar.

shared/src/main/java/com/faforever/neroxis/biomes/Biomes.java
Sheikah45: That is fine with me

Sheikah45:
> We should probably replace the png with a dds. I noticed that the time it takes to compress the dds file is even longer than my machine takes for the whole map generation of a 20km map. The only open question I see is, how do we keep it editable for the future. Every time you modify a dds and resave it, the compression degrades it further. So we need an original png file somewhere. That doesn't have to be shipped in the jar.

How big is it as an uncompressed dds?

shared/src/main/java/com/faforever/neroxis/util/serial/biome/TerrainMaterials.java
Sheikah45: Just initialize the array as new String[TERRAIN_NORMAL_COUNT]; like the others

BlackYps:
Uncompressed is 16 MB without mipmaps, 21.3 MB with mipmaps. Compressed is a quarter of these values, so 4 and 5.3

shared/src/main/java/com/faforever/neroxis/util/serial/biome/TerrainMaterials.java
BlackYps: Right now it is always 10, but it actually doesn't have to be. It could be any other number as well. There is no requirement for it to be 10. Does it still make sense to initialize it with a fixed length then?

Sheikah45: No but then the comment doesn't really make sense either.

BlackYps: yeah, true. Did only think this through after I pushed

BlackYps: So what do we do?

Sheikah45:
The best thing then would probably be to add the compression to dds into the build process and exclude the png from the jar

shared/src/main/java/com/faforever/neroxis/util/serial/biome/TerrainMaterials.java
Sheikah45: Then you could just leave it as it was.

BlackYps:
should the png then be saved in a different location or does that not matter?

Sheikah45:
It could be moved somewhere else or it can just be excluded from the jar using gradle config

BlackYps:
I added the dds, but I don't know how to exclude the png from the jar

Sheikah45:
I excluded it and also fixed the dds export so that it would work when run from a jar and not just from source

Sheikah45:
Although it looks like determinism was broken somehow somewhere according to the failing test

Sheikah45:
Loki was still referenced in some places

generator/src/main/java/com/faforever/neroxis/generator/texture/PbrTextureGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+        return ParameterConstraints.builder()
+                .biomes("Sunset")
+                .build();
+    }
+
+    @Override
+    protected void setupTexturePipeline() {
+        BooleanMask debris = slope.copyAsBooleanMask(.2f);
+        BooleanMask cliff = slope.copyAsBooleanMask(.75f);
+        BooleanMask extendedCliff = slope.copyAsBooleanMask(.75f).inflate(2f);
```
Sheikah45: you should just copy cliff for extended cliff

generator/src/main/java/com/faforever/neroxis/generator/texture/PbrTextureGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+        FloatMask waterBeach = realWater.copy().inflate(11)
+                .subtract(realPlateaus)
+                .subtract(slope.copyAsBooleanMask(.05f).invert())
+                .add(realWater)
+                .copyAsFloatMask(0, 1)
+                .blur(4);
+        FloatMask cliffMask = cliff.copyAsFloatMask(0, 1).blur(4).add(cliff, 1f).blur(2).add(cliff, 0.5f);
+
+        BooleanMask shadowsInWater = shadowsMask.copy().multiply(realWater.copy().setSize(map.getSize()));
+        shadows.setToValue(shadowsInWater.copy(), 1f);
```
Sheikah45: you don't need the copy here

generator/src/main/java/com/faforever/neroxis/generator/texture/PbrTextureGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+        shadowsInWater.add(realLand.copy().setSize(map.getSize()), shadowsInWater.copy().inflate(6));
+        shadows.subtract(realWater.copy().setSize(map.getSize()),
```
Sheikah45: multiple places you set realLand and realWater back to map.getSize so they should be consolidated rather than performing the same operation over again

Sheikah45 (changes requested):
Just a few more small things in the pbr texture generator for efficiency

generator/src/main/java/com/faforever/neroxis/generator/texture/PbrTextureGenerator.java
BlackYps: hm, I only see a double resizing of realWater. Not sure if I can consolidate the rest

BlackYps:
Don't merge yet, btw. I still want to tune some water settings

generator/src/main/java/com/faforever/neroxis/generator/texture/PbrTextureGenerator.java
Sheikah45: You are correct I misread realLand somehwere else

BlackYps:
This is finished now.
It does rely on fafdevelop, so a release should wait until the next game patch.

Sheikah45 (dismissed):

Sheikah45:
What is the status of this now?

BlackYps:
The fa release happened, so that is set. I just need to resolve the conflicts now.
And we should probably make a client release with the ability to select the biome

Sheikah45 (approved):

PR #373, opened 2023-11-07
Improve pbr generator

BlackYps:
Adds a texture size option and better error handling

### DIFF
--- a/toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
+++ b/toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
@@ -26,21 +26,36 @@ public class PbrTextureGenerator implements Callable<Integer> {
     private CommandLine.Model.CommandSpec spec;
     @CommandLine.Mixin
     private OutputFolderMixin outputFolderMixin;
+    @CommandLine.Option(names = "--size", defaultValue = "1024", description = "Size of the input textures in pixels. Defaults to 1024.")
+    private Integer textureImageSize;
 
     @Override
     public Integer call() throws Exception {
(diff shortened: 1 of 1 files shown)


toolsuite/src/main/java/com/faforever/neroxis/toolsuite/PbrTextureGenerator.java
```diff
@@ -26,21 +26,36 @@ public class PbrTextureGenerator implements Callable<Integer> {
...
     @CommandLine.Mixin
     private OutputFolderMixin outputFolderMixin;
+    @CommandLine.Option(names = "--size", defaultValue = "1024", description = "Size of the input textures in pixels. Defaults to 1024.")
+    private Integer textureImageSize;
 
     @Override
     public Integer call() throws Exception {
+        if (!isPowerOfTwo(textureImageSize)) {
+            throw new IllegalArgumentException("Texture size must be a power of two!");
+        }
```
Sheikah45: Rather than checking on call you should make a setter and annotate the setter with the option and perform the check there.

BlackYps: Why is that better?

Sheikah45: Because then you can throw the picocli parameter exception and it can print out the applicable help

Sheikah45: Also then we will know the value is never an invalid value

BlackYps: The value could never be an invalid value even with the old implementation, but I changed it

PR #379, opened 2023-12-05
Change shader back to the original

BlackYps:


shared/src/main/java/com/faforever/neroxis/exporter/MapExporter.java
```diff
@@ -6,12 +6,25 @@
...
             Files.createDirectories(mapPath);
 
+            if (map.getTerrainShaderPath().equals(LEGACY_SHADER_NAME)) {
+                if (map.getCompressedNormal() != null) {
+                    SCMapExporter.exportNormals(mapPath, map);
+                }
+                if (map.getCompressedShadows() != null) {
+                    SCMapExporter.exportShadows(mapPath, map);
+                }
+            } else {
```
Sheikah45: This should probably be an else if

Sheikah45 (approved):

PR #382, opened 2023-12-12
Enable querying the biome list

BlackYps:
Now the client can use this to provide an option to specify the biome that should be generated

### DIFF
--- a/generator/src/main/java/com/faforever/neroxis/generator/Biome.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/Biome.java
@@ -1,0 +1,32 @@
+package com.faforever.neroxis.generator;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public enum Biome {
+    BRIMSTONE("Brimstone"),
+    DESERT("Desert"),
+    EARLYAUTUMN("EarlyAutumn"),
+    FRITHEN("Frithen"),
+    LOKI("Loki"),
+    MARS("Mars"),
+    MOONLIGHT("Moonlight"),
+    PRAYER("Prayer"),
+    STONES("Stones"),
+    SYRTIS("Syrtis"),
+    WINDINGRIVER("WindingRiver"),
+    WONDER("Wonder");
+
+    private final String value;
+
+    public static int getIndexByValue(String value) {
+        for (int i = 0; i < values().length; i++) {
+            if (values()[i].getValue().equals(value)) {
+                return i;
+            }
+        }
+        return -1;
+    }
+}
--- a/generator/src/main/java/com/faforever/neroxis/generator/Biome.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/Biome.java
@@ -20,13 +20,4 @@ public enum Biome {
     WONDER("Wonder");
 
     private final String value;
-
-    public static int getIndexByValue(String value) {
-        for (int i = 0; i < values().length; i++) {
-            if (values()[i].getValue().equals(value)) {
-                return i;
-            }
-        }
-        return -1;
-    }
 }
--- a/generator/src/main/java/com/faforever/neroxis/generator/Biome.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/Biome.java
@@ -20,4 +20,13 @@ public enum Biome {
     WONDER("Wonder");
 
     private final String value;
+
+    public static Biome getByValue(String value) {
+        for (Biome biome : values()) {
+            if (biome.getValue().equalsIgnoreCase(value)) {
+                return biome;
+            }
+        }
+        return null;
+    }
 }
--- a/generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
@@ -121,6 +121,12 @@ public class MapGenerator implements Callable<Integer> {
         System.out.println(Arrays.stream(MapStyle.values()).map(MapStyle::toString).collect(Collectors.joining("\n")));
     }
 
+    @Command(name = "biomes", aliases = {
+            "--biomes"}, description = "Prints the biomes available", versionProvider = VersionProvider.class, usageHelpAutoWidth = true)
+    private void printBiomes() {
+        System.out.println(Arrays.stream(Biome.values()).map(Biome::getValue).collect(Collectors.joining("\n")));
+    }
+
     @Override
     public Integer call() throws Exception {
         Locale.setDefault(Locale.ROOT);
@@ -213,7 +219,7 @@ public class MapGenerator implements Callable<Integer> {
         ParameterOptions parameterOptions = tuningOptions.getParameterOptions();
         if (parameterOptions != null) {
             if (parameterOptions.getBiome() != null) {
-                generatorParametersBuilder.biome(parameterOptions.getBiome());
+                generatorParametersBuilder.biome(Biomes.loadBiome(parameterOptions.getBiome()));
             }
             if (parameterOptions.getLandDensity() != null) {
                 generatorParametersBuilder.landDensity(parameterOptions.getLandDensity());
@@ -367,7 +373,7 @@ public class MapGenerator implements Callable<Integer> {
         randomizeOptions(generatorParametersBuilder);
 
         if (optionBytes.length == 11) {
-            generatorParametersBuilder.biome(Biomes.loadBiome(Biomes.BIOMES_LIST.get(optionBytes[3])));
+            generatorParametersBuilder.biome(Biomes.loadBiome(Biome.values()[optionBytes[3]].getValue()));
             generatorParametersBuilder.landDensity(MathUtil.normalizeBin(optionBytes[4], NUM_BINS));
             generatorParametersBuilder.plateauDensity(MathUtil.normalizeBin(optionBytes[5], NUM_BINS));
             generatorParametersBuilder.mountainDensity(MathUtil.normalizeBin(optionBytes[6], NUM_BINS));
@@ -395,7 +401,7 @@ public class MapGenerator implements Callable<Integer> {
         generatorParametersBuilder.reclaimDensity(MathUtil.discretePercentage(random.nextFloat(), NUM_BINS));
         generatorParametersBuilder.mexDensity(MathUtil.discretePercentage(random.nextFloat(), NUM_BINS));
         generatorParametersBuilder.biome(
-                Biomes.loadBiome(Biomes.BIOMES_LIST.get(random.nextInt(Biomes.BIOMES_LIST.size()))));
+                Biomes.loadBiome(Biome.values()[random.nextInt(Biome.values().length)].getValue()));
         generatorParametersBuilder.terrainSymmetry(getValidTerrainSymmetry());
     }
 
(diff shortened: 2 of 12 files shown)


generator/src/main/java/com/faforever/neroxis/generator/cli/ParameterOptions.java
```diff
@@ -23,7 +21,7 @@ public class ParameterOptions {
     private Float reclaimDensity;
     private Float mexDensity;
     private Symmetry terrainSymmetry;
-    private Biome biome;
+    private String biome;
```
Sheikah45: This should still be a biome not a string.

generator/src/main/java/com/faforever/neroxis/generator/cli/ParameterOptions.java
```diff
@@ -62,6 +60,6 @@ public void setTerrainSymmetry(Symmetry terrainSymmetry) {
 
     @Option(names = "--biome", order = 8, description = "Texture biome for the generated map")
     public void setBiome(String biome) {
-        this.biome = Biomes.loadBiome(biome);
+        this.biome = biome;
```
Sheikah45: We can still do the loading here

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
```diff
@@ -413,7 +419,7 @@ private void encodeMapName() {
                 optionArray = new byte[]{(byte) generatorParameters.spawnCount(),
                                          (byte) (generatorParameters.mapSize() / 64),
                                          (byte) generatorParameters.numTeams(),
-                                         (byte) Biomes.BIOMES_LIST.indexOf(generatorParameters.biome().name()),
+                                         (byte) Biome.getIndexByValue(generatorParameters.biome().name()),
```
Sheikah45: biome.ordinal()

generator/src/main/java/com/faforever/neroxis/generator/Biome.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private final String value;
+
+    public static int getIndexByValue(String value) {
+        for (int i = 0; i < values().length; i++) {
+            if (values()[i].getValue().equals(value)) {
+                return i;
+            }
+        }
+        return -1;
```
Sheikah45: This shouldn't be needed because we should just call ordinal on the biome object

Sheikah45 (changes requested):

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
BlackYps: how? The generatorParameters has the actual biome object. The one that stores all the biome settings. And it needs to stay that way because the generatorParameters get referenced in a lot of places to retrieve the actual biome settings.
To get the ordinal we need the enum.

BlackYps: If we store the enum in the parameterOptions then we could retrieve it from there, would that work?

Sheikah45: Yes sorry I missed it wasn't the enum. It should just be the enum

BlackYps: Should I maybe rename one of the biome classes to make it less confusing?

BlackYps: When I try to put the enum in the ParameterOptions then I still need a helper method to retrieve the right enum. And the tests showed another problem. In the MapGenerator L422 I can't use the tuningOptions to set the biome. These seem to come from the explicitly set command line parameters. When the biome is not specified it gets randomly chosen when intializing the generatorParameters

Sheikah45: You don't need a helper method. You can set the biome enum as the input to the option set method like we do with map styles. And I would get rid of the whole value field of the enum and just use the enum name itself

generator/src/main/java/com/faforever/neroxis/generator/Biome.java
```diff
@@ -0,0 +1,32 @@
+package com.faforever.neroxis.generator;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public enum Biome {
```
Sheikah45: This name needs to be different otherwise it is confusing having two classes named Biome

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
Sheikah45: As for the second point, yes you have to use the generatorParameters so we can either put the biome enum  into the generator parameters and delay loading or add the enum to the Biome object class and retrieve it from there.

BlackYps: Is the resource path not case sensitive? The convention for enum is to be all upper case, but the folders are not.

BlackYps: ping

Sheikah45: We can set them upper base or just convert the enum

BlackYps: how would converting the enum work?

Sheikah45: By just using a capitalize method that takes an enum as a parameter or some enum to folder name method

shared/src/main/java/com/faforever/neroxis/biomes/Biomes.java
```diff
@@ -20,9 +20,6 @@ public class Biomes {
     // ├-- props.json <required>
     // ├-- WaterSettings.scmwtr <required>
     // └-- Light.scmlighting <required>
-    public static final List<String> BIOMES_LIST = List.of("Brimstone", "Desert", "EarlyAutumn", "Frithen",
-                                                           "Loki", "Mars", "Moonlight", "Prayer", "Stones",
-                                                           "Syrtis", "WindingRiver", "Wonder");
     private static final String CUSTOM_BIOMES_DIR = "/custom_biome/";
 
     public static Biome loadBiome(String folderPath) {
```
Sheikah45: I would change this to just use the enum

shared/src/main/java/com/faforever/neroxis/biomes/Biomes.java
```diff
@@ -78,7 +75,7 @@ public static Biome loadBiome(String folderPath) {
                     String.format("An error occurred while loading %s LightingSettings\n", folderPath), e);
         }
 
-        return new Biome(terrainMaterials.getName(), terrainMaterials, propMaterials, decalMaterials, waterSettings,
+        return new Biome(BiomeName.getByValue(terrainMaterials.getName()), terrainMaterials, propMaterials, decalMaterials, waterSettings,
```
Sheikah45: You can just use the passed in enum if changed as suggested

generator/src/main/java/com/faforever/neroxis/generator/cli/ParameterOptions.java
```diff
@@ -61,7 +60,7 @@ public void setTerrainSymmetry(Symmetry terrainSymmetry) {
     }
 
     @Option(names = "--biome", order = 8, description = "Texture biome for the generated map")
-    public void setBiome(String biome) {
-        this.biome = Biomes.loadBiome(biome);
+    public void setBiomeName(String biome) {
+        this.biomeName = BiomeName.getByValue(biome);
```
Sheikah45: This should be the actual enum as the parameter. Then you can use the completion candidates like for symmetry above

shared/src/main/java/com/faforever/neroxis/importer/SCMapImporter.java
```diff
@@ -198,7 +199,7 @@ public static SCMap importSCMAP(Path folderPath) throws IOException {
         in.close();
 
         SCMap map = new SCMap(widthInt,
-                              new Biome("loaded", mapTerrainMaterials, new PropMaterials(), new DecalMaterials(),
+                              new Biome(BiomeName.getByValue("loaded"), mapTerrainMaterials, new PropMaterials(), new DecalMaterials(),
```
Sheikah45: This doesn't really work anymore so should just be passed directly as null

Sheikah45 (changes requested):

shared/src/main/java/com/faforever/neroxis/biomes/Biomes.java
BlackYps: Change what exactly? The list of biomes is now in its separate enum class. Before it was here because it wasn't in a separate class, so now it is deleted from here.

BlackYps: Ah, I see now you meant the string parameter. I first thought you were commenting on the deleted lines

shared/src/main/java/com/faforever/neroxis/biomes/Biomes.java
BlackYps: the terrainmaterials get compiled from JSON. Does using enums work there without any further changes?

BlackYps: Does it even make sense that the terrainmaterials specify the name again? It seems redundant

BlackYps: Ah I see, I just take the parameter from the very top of the function

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
```diff
@@ -121,6 +122,12 @@ private void printStyles() {
         System.out.println(Arrays.stream(MapStyle.values()).map(MapStyle::toString).collect(Collectors.joining("\n")));
     }
 
+    @Command(name = "biomes", aliases = {
+            "--biomes"}, description = "Prints the biomes available", versionProvider = VersionProvider.class, usageHelpAutoWidth = true)
+    private void printBiomes() {
+        System.out.println(Arrays.stream(BiomeName.values()).map(BiomeName::getValue).collect(Collectors.joining("\n")));
```
Sheikah45: This should us to string and not getValue

shared/src/main/java/com/faforever/neroxis/biomes/BiomeName.java
```diff
@@ -0,0 +1,23 @@
...
+    LOKI("Loki"),
+    MARS("Mars"),
+    MOONLIGHT("Moonlight"),
+    PRAYER("Prayer"),
+    STONES("Stones"),
+    SYRTIS("Syrtis"),
+    WINDINGRIVER("WindingRiver"),
+    WONDER("Wonder");
+
+    private final String value;
```
Sheikah45: This should probably be named better so that it is understood what it is used for

Sheikah45 (changes requested):

shared/src/main/java/com/faforever/neroxis/biomes/BiomeName.java
BlackYps: `folderName` maybe?

Sheikah45: That works

Sheikah45 (approved):

PR #390, opened 2024-04-25
Change light multiplier to not trigger the advanced water shader

BlackYps:
Closes #389

PR #397, opened 2024-05-27
Fix previews

BlackYps:
#390 had the unintended side-effect that the water in the previews looked really bad.

after #390:
neroxis_map_generator_snapshot_rs3yp2lri4hec_aqeaeaqiaeaqk_preview

this branch:
neroxis_map_generator_snapshot_yw3j5jzjesrgw_aqiaebqip5xeo4aahm_preview


Sheikah45 (commented):
sunIllumination is not the proper name. The dot product effectively represents how much light from the sun angle makes it back to the camera position of the preview.

BlackYps:
I changed it from reflectance because I first thought it was refering to some material property of the terrain. But what this mask actually stores is how bright the terrain is because of the illumination of the sun. We assume a diffuse reflection here (because there is no code to calculate specular reflection).

BlackYps:
We could also call it irradiance

Sheikah45:
That would work

PR #398, opened 2024-06-01
Add boulder field prop generator

BlackYps:
Adds the changes to the HIGH_RECLAIM generator from #384 as a new prop generator.
grafik

It has rather high reclaim, but not as high as the HIGH_RECLAIM generator. We could add this as a second prop generator to the HIGH_RECLAIM map style or leave leave that style as is.


Sheikah45:
We can add it as an option to the High reclaim style

Sheikah45 (approved):

PR #400, opened 2024-06-17
Odd symmetry fixes

clivepaterson:
So after much digging and debugging, I found the main cause of the issue with Odd Symmetry mapgens.
I basically affects functions that use `applyWithSymmetry()`

And the reason is because points can land on the same pixel after rotation:
after_rotation

And then some points don't get filled in, and are missed.

So I've played with some strategies to resolve this issue, and so far it's looking pretty decent.

### DIFF
--- a/shared/src/main/java/com/faforever/neroxis/mask/BooleanMask.java
+++ b/shared/src/main/java/com/faforever/neroxis/mask/BooleanMask.java
@@ -82,7 +82,7 @@ public final class BooleanMask extends PrimitiveMask<Boolean, BooleanMask> {
         enqueue(dependencies -> {
             T source = (T) dependencies.get(0);
             apply((x, y) -> setPrimitive(x, y, source.valueAtGreaterThanEqualTo(x, y, minValue)
-                    && source.valueAtLessThanEqualTo(x, y, maxValue)));
+                                               && source.valueAtLessThanEqualTo(x, y, maxValue)));
         }, other);
     }
 
@@ -746,12 +746,12 @@ public final class BooleanMask extends PrimitiveMask<Boolean, BooleanMask> {
                 Vector2 previousLoc = checkPoints.get(checkPoints.size() - 1);
                 float angle = (float) ((random.nextFloat() - .5f) * 2 * StrictMath.PI / 2f) + previousLoc.angleTo(end);
                 if (symmetrySettings.terrainSymmetry() == Symmetry.POINT4
-                        && angle % (StrictMath.PI / 2) < StrictMath.PI / 8) {
+                    && angle % (StrictMath.PI / 2) < StrictMath.PI / 8) {
                     angle += (random.nextBoolean() ? -1 : 1) * (random.nextFloat() * .5f + .5f) * 2f * StrictMath.PI
-                            / 4f;
+                             / 4f;
                 }
                 float magnitude = random.nextFloat() * (midPointMaxDistance - midPointMinDistance)
-                        + midPointMinDistance;
+                                  + midPointMinDistance;
                 Vector2 nextLoc = new Vector2(previousLoc).addPolar(angle, magnitude);
                 checkPoints.add(nextLoc);
             }
@@ -770,8 +770,8 @@ public final class BooleanMask extends PrimitiveMask<Boolean, BooleanMask> {
                     }
                     float magnitude = StrictMath.max(1, random.nextFloat() * maxStepSize);
                     float angle = oldAngle * .5f
-                            + location.angleTo(nextLoc) * .5f
-                            + (random.nextFloat() - .5f) * 2f * maxAngleError;
+                                  + location.angleTo(nextLoc) * .5f
+                                  + (random.nextFloat() - .5f) * 2f * maxAngleError;
                     location.addPolar(angle, magnitude).round();
                     oldAngle = angle;
                     numSteps++;
@@ -927,8 +927,8 @@ public final class BooleanMask extends PrimitiveMask<Boolean, BooleanMask> {
             for (int y2 = minY; y2 < maxY; ++y2) {
                 int bitIndex = bitIndex(x2, y2, getSize());
                 if (inBounds(x2, y2)
-                        && getBit(bitIndex, maskCopy) != value
-                        && (x - x2) * (x - x2) + (y - y2) * (y - y2) <= radius2) {
+                    && getBit(bitIndex, maskCopy) != value
+                    && (x - x2) * (x - x2) + (y - y2) * (y - y2) <= radius2) {
                     setBit(bitIndex, value, maskCopy);
                 }
             }
@@ -1007,10 +1007,10 @@ public final class BooleanMask extends PrimitiveMask<Boolean, BooleanMask> {
         return ((x > 0 && getPrimitive(x - 1, y) != value)
                 || (y > 0 && getPrimitive(x, y - 1) != value)
                 || (x
-                < size - 1
-                && getPrimitive(x
-                                        + 1, y)
-                != value)
+                    < size - 1
+                    && getPrimitive(x
+                                    + 1, y)
+                       != value)
                 || (y < size - 1 && getPrimitive(x, y + 1) != value));
     }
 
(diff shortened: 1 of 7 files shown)


clivepaterson:
Some before and afters:

Before:
image

After:
image

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -237,7 +287,11 @@ public FloatMask addGaussianNoise(float scale) {
      * @param scale Multiplicative factor for the noise
      */
     public FloatMask addWhiteNoise(float scale) {
-        return addPrimitiveWithSymmetry(SymmetryType.SPAWN, (x, y) -> random.nextFloat() * scale);
+        if (symmetrySettings.getSymmetry(SymmetryType.SPAWN).isOddSymmetry()) {
```
clivepaterson, line 290: Before:
image

After:
image

I tried to figure how to make this perfect symmetry without any artifacting, but found it impossible.. lol..
So I kinda gave up... and just made it look uniform.

shared/src/main/java/com/faforever/neroxis/mask/BooleanMask.java
```diff
@@ -1008,11 +1028,21 @@ public BooleanMask dilute(float strength, int count) {
...
-                        applyAtSymmetryPoints(x, y, symmetryType, (sx, sy) -> setBit(sx, sy, true, size, maskCopy));
-                    }
-                });
+
+                if (symmetrySettings.getSymmetry(SymmetryType.SPAWN).isOddSymmetry()) {
+                    apply((x, y) -> {
+                        if (!getPrimitive(x, y) && random.nextFloat() < strength && isEdge(x, y)) {
+                            setBit(x, y, true, size, maskCopy);
+                        }
+                    });
```
Sheikah45, line 1037: This ends up breaking the symmetry guarantees because a new random is used for every x,y

shared/src/main/java/com/faforever/neroxis/mask/BooleanMask.java
```diff
@@ -1030,11 +1060,19 @@ public BooleanMask erode(float strength, int count) {
...
-                    if (getPrimitive(x, y) && random.nextFloat() < strength && isEdge(x, y)) {
-                        applyAtSymmetryPoints(x, y, symmetryType, (sx, sy) -> setBit(sx, sy, false, size, maskCopy));
-                    }
-                });
+                if (symmetrySettings.getSymmetry(SymmetryType.SPAWN).isOddSymmetry()) {
+                    apply((x, y) -> {
+                        if (getPrimitive(x, y) && random.nextFloat() < strength && isEdge(x, y)) {
+                            setBit(x, y, false, size, maskCopy);
+                        }
+                    });
```
Sheikah45, line 1068: This ends up breaking the symmetry guarantees because a new random is used for every x,y

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
Sheikah45: Unfortunately the rest of the code and generation relies on the symmetry being maintained so just a uniform distribution is not good enough.

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -171,7 +181,43 @@ public FloatMask addPerlinNoise(int resolution, float scale) {
...
+                                numAround++;
+                            }
+                            if (numAround > 0) {
+                                noise.mask[i][j] = total / numAround;
+                            }
+                        }
+                    }
+                }
+                // Free up the memory
+                noise.hasSet = null;
```
Sheikah45, line 219: I don't know what this is doing because noise.hasSet never is set to true anywhere

Sheikah45, line 219: Unfortunately this doesn't free up the memory right away so can result in heavier load on the users computer

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -228,7 +274,11 @@ Vector3 calculateNormalAt(int x, int y, float scale) {
      * @param scale Multiplicative factor for the noise
      */
     public FloatMask addGaussianNoise(float scale) {
-        return addPrimitiveWithSymmetry(SymmetryType.SPAWN, (x, y) -> (float) random.nextGaussian() * scale);
+        if (symmetrySettings.getSymmetry(SymmetryType.SPAWN).isOddSymmetry()) {
+            return apply((x, y) -> addPrimitiveAt(x, y, (float) random.nextGaussian() * scale));
+        } else {
+            return addPrimitiveWithSymmetry(SymmetryType.SPAWN, (x, y) -> (float) random.nextGaussian() * scale);
+        }
```
Sheikah45, line 281: This ends up breaking the symmetry guarantees because a new random is used for every x,y

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -248,7 +302,11 @@ public FloatMask addWhiteNoise(float scale) {
      */
     public FloatMask addWhiteNoise(float minValue, float maxValue) {
         float range = maxValue - minValue;
-        return addPrimitiveWithSymmetry(SymmetryType.SPAWN, (x, y) -> random.nextFloat() * range + minValue);
+        if (symmetrySettings.getSymmetry(SymmetryType.SPAWN).isOddSymmetry()) {
+            return apply((x, y) -> addPrimitiveAt(x, y, random.nextFloat() * range + minValue));
+        } else {
+            return addPrimitiveWithSymmetry(SymmetryType.SPAWN, (x, y) -> random.nextFloat() * range + minValue);
+        }
```
Sheikah45, line 309: This ends up breaking the symmetry guarantees because a new random is used for every x,y

Sheikah45 (changes requested):
Hey clivepaterson, thanks for the contribution. After taking a look through the changes there are a few things that aren't going to work out in the long run. 

I saw that in most places you just added an isOddSymmetry check and if that returned true ended up ignoring the symmetry requirements. Unfortunately this isn't really tractable since the symmetry is required for things like placing spawns or mexes. If it is ignored then the mex placer and spawn placer could end up placing these on invalid locations making them unbuildable for some players.

Additionally I saw you added an if statement to a terrain generator with some hefty logic differences in the body. The terrain generators can have their setup methods overridden or new generators that don't inherit from basic may come about making it hard to keep up with. Additionally the logic heavy if statements can end up making the generation unpredictable and hard to tweak. Resolving this is actually why the styles was used and introduced.

Also adding the hasSet boolean array to every mask as a property is a bit heavy handed. boolean is actually represented by 8 bits so can add a lot to the memory footprint of a mask. This is why the boolean mask uses longs to represent the booleans instead.

I think overall the approach to solving this problem is not to avoid applyWithSymmetry but instead change that function and forceSymmetry so that it can handle the offset rotations.

clivepaterson:
So yes this is a quite a difficult challenge.

The traditional algorthims for pixel rotation are not gonna help much, mentioned in this article:
https://gamedev.stackexchange.com/questions/135091/how-can-i-rotate-pixel-art-sprites-without-the-aesthetics-getting-ruined
The RotSprite method is also impractical. So traditional rotation algorithms aren't overly useful.

There is an even crazier idea:
detecting when a rotate pixel lands on occupied pixel, and move it to an adjacent pixel that hasn't been set.
And I've actually tried this, but the results are not good:
image


So another idea I've been thinking about is to:
1) draw the whole first slice
2) loop through the remaining pixels, and find their correct value by rotating backwards, to lookup the value in the original slice

It's kinda a reversal way of copying the slice, but is means that every pixel will get a value.

Of course this would only work for certain functions like Noise, Perlin, Dilute etc.
And it wouldn't work for mex or spawn placements etc, which are a bit more immune from this rotation artifacting.

Sheikah45:
Hmm yeah your reversal of the symmetry point could work. Although would have to double check what that does to the performance.

Sheikah45:
It would also remove a lot of the if statements that the current solution has. Ideally a solution doesn't really break out of the applySymmetry function.

clivepaterson:
Yep, I'm thinking I'll have a go at the reverse lookup strategy.
And create a new branch if it works.
I should be able to stay within the applySymmetry function too.

I can't see this current branch merging. But it has been useful discussion. Not sure if it's worth keeping the PR up.

I can see myself taking about 1-2 weeks to try the reverse lookup strategy, although it might be fairly easy once I get stuck into it.

For performance I'm guessing it'll be the same, because it should be only 1 rotation calculation per pixel still...
But yeah, the rotation is likely one of the more expensive functions in the code.
Can see the number of ops for sine and cosine here: https://www.agner.org/optimize/instruction_tables.pdf

clivepaterson:
Closed because a better method was found.

PR #412, opened 2024-08-01
Allow visual debugging from any thread

clivepaterson:
This is a small change to make the Visual Debugger work when generating a map.
When doing Swing UI updates, it must be run from the UI thread.
So this uses `SwingUtilities.invokeLater()` to make sure UI updates are run from the UI thread.

Which is nice because we can reliably debug the entire map generation if desired:
EG: this works now without issue:
image

### DIFF
--- a/shared/src/main/java/com/faforever/neroxis/visualization/VisualDebugger.java
+++ b/shared/src/main/java/com/faforever/neroxis/visualization/VisualDebugger.java
@@ -20,10 +20,18 @@ public class VisualDebugger {
     }
 
     public static void visualizeMask(Mask<?, ?> mask, String method, String line) {
-        createGui();
-        String name = mask.getVisualName();
-        name = name == null ? mask.getName() : name;
-        updateList(name + " " + method + " " + line, mask.immutableCopy());
+        Mask<?, ?> copyOfmask = mask.copy();
+        SwingUtilities.invokeLater(
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        createGui();
+                        String name = copyOfmask.getVisualName();
+                        name = name == null ? copyOfmask.getName() : name;
+                        updateList(name + " " + method + " " + line, copyOfmask.immutableCopy());
+                    }
+                }
+        );
     }
 
     public static void createGui() {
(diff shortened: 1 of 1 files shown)


Sheikah45:
Sorry  missed the notification on this. This is actually interesting I wasn't aware that the UI updates had to be from a specific thread. I am curious now how the visual debugger worked at all previously now haha

shared/src/main/java/com/faforever/neroxis/visualization/VisualDebugger.java
```diff
@@ -20,10 +20,18 @@ public static void visualizeMask(Mask<?, ?> mask, String method) {
...
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        createGui();
+                        String name = copyOfmask.getVisualName();
+                        name = name == null ? copyOfmask.getName() : name;
+                        updateList(name + " " + method + " " + line, copyOfmask.immutableCopy());
+                    }
+                }
+        );
```
Sheikah45:
```suggestion
        copyOfmask = mask.immutableCopy()
        SwingUtilities.invokeLater(() -> {
                        createGui();
                        String name = copyOfmask.getVisualName();
                        name = name == null ? copyOfmask.getName() : name;
                        updateList(name + " " + method + " " + line, copyOfmask);
                    });
```
We can use the lambda syntax and also can just use the copy directly rather than create another copy in the lambda

Sheikah45 (changes requested):

clivepaterson:
> Sorry missed the notification on this. This is actually interesting I wasn't aware that the UI updates had to be from a specific thread. I am curious now how the visual debugger worked at all previously now haha

It's an interesting one... TestingGround was alway working fine.
It was just the full map generation having issues, specially when you try multiple Mask's with .startVisualDebug()

I've only come accross this once before many years ago when I was making a Plugin for Outlook using C#...
I was running into this UI thread challenge, I think in that case I solved it with 'Thread.join()'  to get back onto the main UI thread or something...

But I think SwingUtilities.invokeLater() is an accepted / conventient Java/Swing way to get code running back on the UI thread.

Sheikah45:
I think there is a semi colon missing. Once added and tests pass I can merge.

Sheikah45 (approved):

Sheikah45:
Ah yes this is much nicer know. I had always meant to figure out why sometimes the UI bugged out when run with --visualize. Thanks!

PR #414, opened 2024-08-17
New biome inspired by the map Zerg Canyon.

clivepaterson:
New biome idea.
Uses crystals for reclaim and darger textures.

image


shared/src/main/resources/custom_biome/Crystalline/props.json
```diff
@@ -0,0 +1,15 @@
+{
+  "treeGroups": [
+  ],
```
clivepaterson, line 3: What are your feelings about having no tree groups, and no boulders for this biome?

Sheikah45 (approved):

BlackYps:
Very nice to see someone investing the effort to add new biomes! Good work!

shared/src/main/resources/custom_biome/Crystalline/props.json
BlackYps: Having no tree groups is not really an issue, but having no boulders leads to unexpected results, especially with the `boulder field` prop generator. I would suggest moving the long, big crystal to boulders. Then you also get more interesting variation in reclaim placement. At the moment the very big crystal is used excessively and it would look nicer if it gets placed less often (which will happen when it is classified as a boulder).

BlackYps: Actually I already did it here, because it is such a small change https://github.com/FAForever/Neroxis-Map-Generator/pull/416

PR #416, opened 2024-08-18
Move big crystal in crystalline biome to boulder category

BlackYps:


clivepaterson:
Oh yep, makes sense.

I generated a few 10k maps with this change, to see what the mass totals look like.
Most of the maps generated with ~20k worth of props reclaim, which is a good average for 10k maps, (Badlands is about 23k mass which is a good example of an average 10k ladder map).

I did manage to generate 1 map which used the BoulderFieldPropGenerator, and the reclaim total was very high (100k). Maybe too high.
This feels like "The Ditch" kinda map, but then again, "The Ditch" is a classic. I kinda like these high reclaim maps, and I like "The Ditch".

Sheikah45 (approved):

PR #423, opened 2024-10-17
Encode dds as ARGB instead of ABGR

BlackYps:
ABGR is a pretty unusual format. Most software can deal with this just fine, but our unity map editor can't deal with it and reads it as ARGB which causes problems.
Instead of trying to fix the map editor, which would be complicated because unity doesn't even know the ABGR format, we can just output a more standard format here.

### DIFF
--- a/shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
+++ b/shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
@@ -288,25 +288,6 @@ public class ImageUtil {
         return allBytes;
     }
 
-    private static byte[] getRawDDSImageBytes(int size, ByteBuffer imageByteBuffer) {
-        DDSHeader ddsHeader = new DDSHeader();
-        ddsHeader.setWidth(size);
-        ddsHeader.setHeight(size);
-        ddsHeader.setRGBBitCount(32);
-        ddsHeader.setRBitMask(0x000000FF);
-        ddsHeader.setGBitMask(0x0000FF00);
-        ddsHeader.setBBitMask(0x00FF0000);
-        ddsHeader.setABitMask(0xFF000000);
-        ddsHeader.toBytes();
-        byte[] headerBytes = ddsHeader.toBytes();
-        byte[] imageBytes = imageByteBuffer.array();
-        int headerLength = headerBytes.length;
-        int imageLength = size * size * 4;
-        byte[] allBytes = Arrays.copyOf(headerBytes, headerLength + imageLength);
-        System.arraycopy(imageBytes, 0, allBytes, headerLength, imageLength);
-        return allBytes;
-    }
-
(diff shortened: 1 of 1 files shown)


BlackYps:
Actually I'm not sure if the encoding is now ARGB or BGRA, but it doesn't really matter. I tested that the map editor now correctly reads the texture and other software can still read it without problems.

Sheikah45:
What are other maps encoded as? I thought the map gen did it in the same way other maps do

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
```diff
@@ -288,25 +289,6 @@ private static byte[] getCompressedDDSImageBytes(int size, ByteBuffer imageByteB
...
-        ddsHeader.setABitMask(0xFF000000);
-        ddsHeader.toBytes();
-        byte[] headerBytes = ddsHeader.toBytes();
-        byte[] imageBytes = imageByteBuffer.array();
-        int headerLength = headerBytes.length;
-        int imageLength = size * size * 4;
-        byte[] allBytes = Arrays.copyOf(headerBytes, headerLength + imageLength);
-        System.arraycopy(imageBytes, 0, allBytes, headerLength, imageLength);
-        return allBytes;
-    }
```
Sheikah45, line 308: Why was this method deleted?

BlackYps:
what do you mean? This is just about the mapwide texture that the map generator produces. Most maps don't have this texture at all

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
BlackYps: It was unused.

BlackYps: I noticed that this also uses ABGR order. So instead of thinking a lot about if I need to change the byte order there as well I deleted it. Without the method being used I can't really decide which order is appropriate and I didn't want anyone to use it in the future, expecting it to behave similar to `writeRawDDS`

clivepaterson:
Oh yes, I have seen this strange effect in the Mapeditor.
image

I have always wondered what it is...
The first few times I saw it, I though I broke the mapgen code somehow...

Anyway, yep this a good fix... mystery solved.

clivepaterson:
I have gave this branch a test, and generated a map with the SunsetTextureGenerator.

It looked like this for me when I openned it in the MapEditor: (not sure if it's expected)
image

And it looks normal in game:
image

BlackYps:
That upper image should be correct. The sunset biome uses a different shader and in the current release the map editor can't yet display maps correctly that use this shader. You can go to the textures tab and confirm how the texture looks without the lighting applied.

Sheikah45:
> what do you mean? This is just about the mapwide texture that the map generator produces. Most maps don't have this texture at all

Ah didn't realize this was just the map wide

PR #422, opened 2024-10-09
Concept for showing boulders on the mapgen preview.

clivepaterson:
So I played around a bit, with the idea of showing the bolder props on the map preview.
I've made it so that each biome can set a colour to use in the preview.

It's took a lot of tries at different things. Having 1 colour to work well on 50% of maps and bad on others..
Having white on white doesn't work well and dark maps need a fairly bright boulder colour, (having an outline helps).

Anyway, I'll grab screen shots of the different biomes, to see what it looks like atm:
Brimstone:
image
image

Desert:
image
image

EarlyAutumn:
image
image

Frithen:
image
image
(this one doesn't look to great actually, now that I'm taking some screen shots)

Mars:
image
image

Moonlight:
image
image
I noticed that moonlight was returning the Mars generator, not sure if it was intended.
So I brought back the moonlight biome, but I adjusted the 'Light.scmlighting' because it's too dark I think.
I can undo this, and set it back to mars, otherwise... this is how it looks
image

Prayer:
image
image

Stones:
image
image

Sunset:
image
image

Syrtis:
image
image

WindingRiver:
image
image

Wonder:
image
image

And last, but not least, Crystalline:
image
image
Which also has some scope creep, to enable the 50 and 150 mass boudlers.
So it looks like this now:
image

Geeeees, so many biomes.....
Anyway's, I found that a different colour per biome gave the most flexibility, however I feel that some are more obvious than others, and some colours at least could use some tweaking...

PR #436, opened 2025-02-24
A new style of terrain generator.

clivepaterson:
5k Rivers:
image

10k Rivers:
image

20k Rivers:
image

5k Rivers And Oceans:
image

10k Rivers And Oceans:
image

20k Rivers And Oceans:
image

This is a 20K Rivers:
image

This is a 20K River And Oceans:
image

### DIFF
--- a/generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
@@ -1,0 +1,12 @@
+package com.faforever.neroxis.generator.resource;
+
+public class RiversAndOceansMexResourceGenerator extends BasicResourceGenerator {
+
+    @Override
+    public void setupPipeline() {
+        resourceMask.init(passableLand).startVisualDebugger();
+        resourceMask.add(passableWater);
+        resourceMask.subtract(unbuildable);
+        waterResourceMask.init(resourceMask);
+    }
+}
--- a/generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
@@ -4,7 +4,7 @@ public class RiversAndOceansMexResourceGenerator extends BasicResourceGenerator 
 
     @Override
     public void setupPipeline() {
-        resourceMask.init(passableLand).startVisualDebugger();
+        resourceMask.init(passableLand);
         resourceMask.add(passableWater);
         resourceMask.subtract(unbuildable);
         waterResourceMask.init(resourceMask);
--- a/generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
@@ -2,7 +2,6 @@ package com.faforever.neroxis.generator.terrain;
 
 import com.faforever.neroxis.brushes.Brushes;
 import com.faforever.neroxis.generator.GeneratorParameters;
-import com.faforever.neroxis.generator.ParameterConstraints;
 import com.faforever.neroxis.map.SCMap;
 import com.faforever.neroxis.map.SymmetrySettings;
 import com.faforever.neroxis.mask.BooleanMask;
@@ -12,18 +11,9 @@ import com.faforever.neroxis.util.vector.Vector3;
 
 public class RiversTerrainGenerator extends BasicTerrainGenerator {
 
-    private BooleanMask riverExclusionMask;
-    private BooleanMask riverMountains;
-    private BooleanMask riverMask;
-
-    protected boolean watermap;
-
-    @Override
-    public ParameterConstraints getParameterConstraints() {
-        return ParameterConstraints.builder()
-                                   .mapSizes(768, 1024)
-                                   .build();
-    }
+    protected BooleanMask riverExclusionMask;
+    protected BooleanMask riverMountains;
+    protected BooleanMask riverMask;
 
     @Override
     public void initialize(SCMap map, long seed, GeneratorParameters generatorParameters,
@@ -41,8 +31,6 @@ public class RiversTerrainGenerator extends BasicTerrainGenerator {
         mountainBrushSize = 24;
         mountainBrushDensity = 8f;
         mountainBrushIntensity = 0.4f;
-
-        watermap = false;
     }
 
     @Override
@@ -54,27 +42,19 @@ public class RiversTerrainGenerator extends BasicTerrainGenerator {
         int riversScale = mapSize / 64;
         FloatMask rivers = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(), "rivers", true);
         rivers.addPerlinNoise(96 + riversScale, 1);
-        riverMask = watermap ?
-                rivers.copyAsBooleanMask(0.2f, 0.8f) :
-                rivers.copyAsBooleanMask(0.5f, 0.65f);
+        riverMask = rivers.copyAsBooleanMask(0.5f, 0.65f);
 
-        var riverExclusion = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(), "riversExclusion", true);
+        FloatMask riverExclusion = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(), "riversExclusion", true);
         riverExclusion.addPerlinNoise(256, 1);
         float exclusionThickness = 0.06f * ((float)mapSize / 256f);
         riverExclusionMask = riverExclusion.copyAsBooleanMask(0.5f, 0.5f + exclusionThickness);
-        if (!watermap) {
-            riverMask.subtract(riverExclusionMask);
-        }
+        riverMask.subtract(riverExclusionMask);
         riverExclusionMask.setSize(mapSize+1);
 
         riverMask.invert();
         riverMask.blur(10);
 
-        // This ensures some land connection's between teams,
-        // For larger water maps we don't need it though.
-        if (!watermap || mapSize < 512) {
-            riverMask.add(connections.copy().dilute(1, 10).setSize(riverMask.getSize()));
-        }
+        riverMask.add(connections.copy().dilute(1, 10).setSize(riverMask.getSize()));
 
         riverMask.erode(0.3f, 10);
 
@@ -105,9 +85,9 @@ public class RiversTerrainGenerator extends BasicTerrainGenerator {
 
         FloatMask plats = new FloatMask(mapSize, getRandom().nextLong(), plateaus.getSymmetrySettings(), "mountainplateaus", true);
         plats.addPerlinNoise(32, 1f);
-        var platMountains = plats.copyAsBooleanMask(plateauDensity);
+        BooleanMask platMountains = plats.copyAsBooleanMask(plateauDensity);
 
-        var plateauExclusion = new BooleanMask(mapSize, random.nextLong(), getSymmetrySettings(), "plateauExclusion", true);
+        BooleanMask plateauExclusion = new BooleanMask(mapSize, random.nextLong(), getSymmetrySettings(), "plateauExclusion", true);
         String[] SPAWN_MASK_BRUSHES = {
                 "mountain4.png",
                 "mountain7.png",
@@ -137,7 +117,7 @@ public class RiversTerrainGenerator extends BasicTerrainGenerator {
         heightmapPlateaus.useBrushWithinAreaWithDensity(plateaus, brush, plateauBrushSize, plateauBrushDensity,
                                                         plateauBrushIntensity, false).clampMax(plateauHeight);
 
-        var inverseRiverMask = riverMask.copy().invert().setSize(heightmapPlateaus.getSize()).blur(6);
+        BooleanMask inverseRiverMask = riverMask.copy().invert().setSize(heightmapPlateaus.getSize()).blur(6);
         heightmapPlateaus
                 .subtract(inverseRiverMask, heightmapPlateaus)
                 .subtract(inverseRiverMask, 0.45f)
@@ -176,7 +156,7 @@ public class RiversTerrainGenerator extends BasicTerrainGenerator {
 
         FloatMask rampExclusion = new FloatMask(ramps.getSize(), random.nextLong(), this.symmetrySettings, "rampExclusion", true);
         rampExclusion.addPerlinNoise(64, 1);
-        var rampExclusionMask = rampExclusion.copyAsBooleanMask(0.4f, 0.8f);
+        BooleanMask rampExclusionMask = rampExclusion.copyAsBooleanMask(0.4f, 0.8f);
 
         ramps.subtract(rampExclusionMask);
     }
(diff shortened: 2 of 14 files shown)


generator/src/main/java/com/faforever/neroxis/generator/texture/MoonlightTextureGenerator.java
```diff
@@ -8,6 +8,6 @@ public class MoonlightTextureGenerator extends LegacyTextureGenerator {
 
     @Override
     public Biome loadBiome() {
-        return Biomes.loadBiome(BiomeName.MOONLIGHT);
+        return Biomes.loadBiome(BiomeName.MARS);
```
clivepaterson: I re-enabled this texture generator back in this PR:
https://github.com/FAForever/Neroxis-Map-Generator/pull/422/files#diff-fec5b5d6e8983a01dee9e9c60f08f9410b46b149c0c14b94a4766f681f561185

But I heard Farm's mention on stream that it has bad lighting, which was kinda my fault when I turned up the Bloom too far on the lighting... lol.

So yep I guess that was a failed experiment, and I should revert this back.

Sheikah45: Why change the biome instead of just disabling the texture generator altogether?

generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
```diff
@@ -0,0 +1,12 @@
+package com.faforever.neroxis.generator.resource;
+
+public class RiversAndOceansMexResourceGenerator extends BasicResourceGenerator {
+
+    @Override
+    public void setupPipeline() {
+        resourceMask.init(passableLand).startVisualDebugger();
```
Sheikah45: Will want to remove the visual debugger call

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,266 @@
...
+        land.setSize(mapSize);
+
+        int riversScale = mapSize / 64;
+        FloatMask rivers = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(), "rivers", true);
+        rivers.addPerlinNoise(96 + riversScale, 1);
+        riverMask = watermap ?
+                rivers.copyAsBooleanMask(0.2f, 0.8f) :
+                rivers.copyAsBooleanMask(0.5f, 0.65f);
+
+        var riverExclusion = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(), "riversExclusion", true);
```
Sheikah45: Style nit of var not being used

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,266 @@
...
+
+        riverMountains = riverMask.copy()
+                                  
+                                  .erode(1f, 5)
+                                  .outline()
+                                  .inflate(2);
+
+        FloatMask riverMountainExclusion = new FloatMask(riverMountains.getSize(), random.nextLong(), this.symmetrySettings, "riverMountainExclusion", true);
+        riverMountainExclusion.addPerlinNoise(128, 1);
+        var riverMountainExclusionMask = riverMountainExclusion.copyAsBooleanMask(0.3f, 0.8f);
```
Sheikah45: Style nit of var

Sheikah45 (changes requested):
I really like the look of the rivers that you were able to achieve.

The only thing I am a little cautious of is the conditionals in the rivers terrain generator. While it seems like a good idea to reduce duplication this was how the generators were done in the past and it lead to a lot of headaches when trying to find tune anything.

This might be a case where DRY leads us astray due to the sensitive nature of the random processes that are done.

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
clivepaterson: Lol, yep. it's a bad habbit from work using C#.
They've all been fixed now.

generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
clivepaterson: Oh, I missed this one.
I removed millions of these calls, but I missed this.

clivepaterson:
Well, build still failing. I'm still trying to figure out why.
In the mean time I made this image.
rivers_collage1_sd

generator/src/main/java/com/faforever/neroxis/generator/texture/MoonlightTextureGenerator.java
clivepaterson: Yep sounds good, that way this file remains intact.

Sheikah45:
The build is failing because for the rivers generator when adding the perlin noise the resolution is too high for some of the map sizes which results in gradient masks of size 1 which then cause issues.

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -12,18 +11,9 @@
...
-    private BooleanMask riverMountains;
-    private BooleanMask riverMask;
-
-    protected boolean watermap;
-
-    @Override
-    public ParameterConstraints getParameterConstraints() {
-        return ParameterConstraints.builder()
-                                   .mapSizes(768, 1024)
-                                   .build();
```
clivepaterson: So I removed this override function for getParameterConstraints() which I thought was causing the unit tests to fail.
But maybe it was some other part of the changes that got the unit test passing.
It's passing for me locally all the time now:
image

Sheikah45: The parameter constraints are just used for selecting the proper generator when there are multiple possibilities. If there is only one of a type it will always be used.

clivepaterson: Yep, I just realised the same thing.
I have replicated the issue and I found out what was causing it:
image

basically the `int resolution` must be less than or equal to the mapSize.

I have pushed  fix.

clivepaterson: I've added a unit test to check all terrain generators with all the various sizes...
image

I'm not sure if it too much testing though, it adds a bit of time to the build.

clivepaterson:
So I might paste the new screen shots of the day...
These are maps that would generate for the 3 v 3 ladder matches:

10k
10k_rivers_3v3

10k
10k_rivers_and_oceans_3v3

15k
15k_rivers_3v3

15k
15k_rivers_and_oceans_3v3

generator/src/test/java/com/faforever/neroxis/generator/MapGeneratorTest.java
```diff
@@ -98,6 +99,14 @@ public void TestParseMapName() {
         assertEquals(generatorParameters.mapSize(), mapSize);
     }
 
+    @ParameterizedTest
+    @ArgumentsSource(ValidTerrainAndMapSizeArgumentProvider.class)
+    public void TestEqualityTerrainAndMapSizeGeneratorSpecified(TerrainStyle terrainStyle, int mapSize) {
+        instance = new MapGenerator();
+
+        new CommandLine(instance).execute("--terrain-style", terrainStyle.toString(), "--map-size", String.valueOf(mapSize));
+    }
```
Sheikah45: This test doesn't test equality for determinism so should just be renamed to like test terrain size compatibility or something

Sheikah45 (changes requested):

generator/src/test/java/com/faforever/neroxis/generator/MapGeneratorTest.java
clivepaterson: Yep, was a badly named test... I've renamed it and added some assertions.

clivepaterson:
And time for another Test / Screenshot of the day.
This one is a 5v5 20k Oceans one:
image

I did experiment with the idea of spawns in the water.
I tried to make it so that players 3 and 4 have a chance of being a water spawn.

But it didn't work out. It was hard to garantee enough mexes and power for the water spawns...

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,228 @@
...
+
+        rampDensity = random.nextFloat() * 0.2f + 0.8f;
+
+        mountainBrushSize = 24;
+        mountainBrushDensity = 8f;
+        mountainBrushIntensity = 0.4f;
+    }
+
+    @Override
+    protected void landSetup() {
```
clivepaterson, line 35: So this is really the hearth of this terrain generator. It uses Perlin noise and copies only values between 0.5 and 0.65 to create a river effect:
image

And then the following is done to transform this into a land mask.

The rivers creates lot of islands due to it's nature, so we create another Perlin noise with thickness and we'll use this to create land paths accross the map:
image

This is the main character of the map, a look at the preview shows how the map turns out:
image

The remaing lines in this function, do this:
Change from rivers mask to land mask:
image

Smooth out the rivers:
image

Garantee pathable land between teams (makes sure river don't cut teams off)
Arguably this should be thicker maybe:
image

Give the river's edges some roughness, othwise the map is very bland:
image
The above has this affect on the rivers edges:
image



Ensure the spawn points have land, otherwise they can have water start, and no mexes...
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,228 @@
...
+            }
+        });
+
+        land.add(riverMask);
+
+        land.setSize(mapSize+1);
+    }
+
+    @Override
+    protected void plateausSetup() {
```
clivepaterson, line 79: This plateusSetup is slightly alternative approach to plateus using Perlin noise and fairly high minValue when converting it to a BooleanMask.
So the results look like this:

Perlin noise:
image

Plateau boolean mask:
image

And then we subtract various exclusion area's like spawn points and rivers and we get this:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,228 @@
...
+        platMountains.setSize(mapSize);
+        plateaus.add(platMountains);
+
+        plateaus.setSize(mapSize + 1);
+        plateaus.subtract(riverMask.copy().invert().blur(10).setSize(plateaus.getSize()));
+    }
+
+    @Override
+    protected void setupPlateauHeightmapPipeline() {
+        String brush = Brushes.GENERATOR_BRUSHES.get(random.nextInt(Brushes.GENERATOR_BRUSHES.size()));
```
clivepaterson, line 112: This version of `setupPlateauHeightmapPipeline()` is bascially the same as the base version with one small difference.

This version subtracts the `riverMask` from the plateau heightmap.
This has the affect of river's that cut through plateeau's.

After plateau brush:
image

After subtracting rivers:
image

And the final heightmap becomes:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,228 @@
...
+
+        heightmapPlateaus
+                .blur(4, plateauBase.copy()
+                                    .inflate(96)
+                                    .subtract(plateauBase.copy().inflate(4)));
+    }
+
+    @Override
+    protected void initRamps() {
+        ramps.setSize(map.getSize() + 1);
```
clivepaterson, line 150: So the default `initRamps()` and `blurRamps()` does mostly weork with this terrain generator, but it does some extra stuff with team connections etc.

So I looked at new approach, use the plateau mask, so that we can put ramps on the edge of plateau's.

So it simply grabs the plateau mask:
image

And uses this cool `outline()` function to find the edge's of all the plateau's:
image

And then we remove random parts, using perlin noise.
So we are left with about 50% of the edges of platea's becoming ramps:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,228 @@
...
+        FloatMask rampExclusion = new FloatMask(ramps.getSize(), random.nextLong(), this.symmetrySettings, "rampExclusion", true);
+        rampExclusion.addPerlinNoise(64, 1);
+        BooleanMask rampExclusionMask = rampExclusion.copyAsBooleanMask(0.4f, 0.8f);
+
+        ramps.subtract(rampExclusionMask);
+    }
+
+    @Override
+    protected void blurRamps() {
+        BooleanMask inflatedRamps = ramps.copy();
```
clivepaterson, line 164: This function has to change due to the nature of this terrain generator.
I found that the base function has `.subtract(noRamps.inflate(4))` which was causing issue's here. So I had to come up with a simple alternative.
So this works pretty well, this is a couple of the steps it does:
image
image
Final result:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,228 @@
...
+                 .blur(4, inflatedRamps.copy().outline().inflate(4))
+                 .blur(8, inflatedRamps.copy().outline().inflate(8))
+                 .blur(4, inflatedRamps.copy().outline().inflate(16))
+                 .clampMin(0f)
+                 .clampMax(255f);
+    }
+
+    @Override
+    protected void mountainSetup() {
+        int mapSize = map.getSize();
```
clivepaterson, line 176: So this `mountainSetup()` a quite different.
I wanted to use the mountain generator along the edges of the rivers.
So that parts of the river's will have mountains alone them.

So to achieve this I used the outline function.

RiverMask:
image

The river's outline:
image

Then we subtract some randomness so that we still have pathable river edges, and we end up with a mounttainMask of this:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -0,0 +1,228 @@
...
+
+        riverMountains.subtract(riverMountainExclusionMask);
+        riverMountains.setSize(mountains.getSize());
+        mountains.add(riverMountains);
+        mountains.setSize(mapSize + 1);
+    }
+
+    @Override
+    protected void spawnTerrainSetup() {
+        int mapSize = map.getSize();
```
clivepaterson, line 198: So this is a slightly simplified version of the `spawnTerrainSetup()` function.

The base version has some code which alter's stuff on the land mask and this was causing all sorts of undesired results.
image

So I've used an override here to skip that part of the base function.

clivepaterson (commented):
So I've had a go at github's self review feautre to see how it works.

And to try give some explanation to all the code that's in this terrain generator, and provide the opportunity for any feedback on any parts of it.

Sheikah45:
I really like the usage of the perlin noise with all of this.

Sheikah45:
@clivepaterson what do you think about my comments about the if statements in the setup path?

clivepaterson:
> @clivepaterson what do you think about my comments about the if statements in the setup path?

I'm trying to remember the comment about the if statements... I though they were resolved by removing the `isOcean` flag with all the `isOcean ? dosomething : orsomethingelse` statments.

But it could be refering to this if statment in RiversTerrainGenerator:landSetup()
image
Which is needed because the `addBrush()` draws the brush for all teams:
image

Or this if statement in RiversAndOceansTerrainGenerator:landSetup())
image
This was intended for the 5k and 7.5k oceans, maps. It help to add a small land connection between the player if there isn't one. But it can be removed.
This is what it does on a 7.5K map (which is currently on 1v1 ladder)
Before Adding Connections:
image
After addings connections:
image
Final Map:
image
And this is an example of a 10k map. Where the connections aren't added.
image
This particular generator works best at generating navy maps at 10K up. I feel like the smaller maps need these land connections to make them more playable / more fun.

And I did a small commit today.. I was generating a bunch of maps, and I was finding a lot of plateus...
And small adjustment has a big affect.
Shown here:
0.83f    (lots of plateus)
image
0.90f    (not much plateus at all)
image
So I added a tiny randomness to the value, which has a really nice affect on the generated maps.

But I think it's good to go.... if the if statements are OK which I think they are.

Sheikah45:
Ah yeah you are right I missed the removal of the isoceans

clivepaterson:
So what you thinking? Is is getting closer to merge time?
Ready to unleash this beast upon the players?

Sheikah45:
Yes I just need to actually press the button.

Sheikah45 (approved):

PR #437, opened 2025-03-14
Accomodate to new shader changes

BlackYps:
I had to rework how the new shaders work in the game. This has implications for the functionality of the toolsuite and the current sunset biome that uses them.
This requires to temporarily disable the sunset biome. If we didn't do this then we would have to release the new map generator and the game patch at exactly the same time, or maps using the sunset biome would be temporarily broken. After the game patch we can enable the biome again.

### DIFF
--- a/shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
+++ b/shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
@@ -206,8 +206,10 @@ public final class FloatMask extends PrimitiveMask<Float, FloatMask> {
 
     Vector3 calculateNormalAt(int x, int y, float scale) {
         float xNormal, yNormal;
-        xNormal = (getPrimitive(x, y) - getPrimitive(x + 1, y)) * scale;
-        yNormal = (getPrimitive(x, y) - getPrimitive(x, y + 1)) * scale;
+        xNormal = ((getPrimitive(x, y) - getPrimitive(x + 1, y)) +
+                  (getPrimitive(x, y + 1) - getPrimitive(x + 1, y + 1))) * 0.5f * scale;
+        yNormal = ((getPrimitive(x, y) - getPrimitive(x, y + 1)) +
+                   (getPrimitive(x + 1, y) - getPrimitive(x + 1, y + 1))) * 0.5f * scale;
         return new Vector3(xNormal, 1, yNormal).normalize();
     }
 
--- a/shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
+++ b/shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
@@ -206,20 +206,8 @@ public final class FloatMask extends PrimitiveMask<Float, FloatMask> {
 
     Vector3 calculateNormalAt(int x, int y, float scale) {
         float xNormal, yNormal;
-        if (x == 0) {
-            xNormal = (getPrimitive(x, y) - getPrimitive(x + 1, y)) * scale;
-        } else if (x == (getSize() - 1)) {
-            xNormal = (getPrimitive(x - 1, y) - getPrimitive(x, y)) * scale;
-        } else {
-            xNormal = (getPrimitive(x - 1, y) - getPrimitive(x + 1, y)) * scale / 2f;
-        }
-        if (y == 0) {
-            yNormal = (getPrimitive(x, y) - getPrimitive(x, y + 1)) * scale;
-        } else if (y == (getSize() - 1)) {
-            yNormal = (getPrimitive(x, y - 1) - getPrimitive(x, y)) * scale;
-        } else {
-            yNormal = (getPrimitive(x, y - 1) - getPrimitive(x, y + 1)) * scale / 2f;
-        }
+        xNormal = (getPrimitive(x, y) - getPrimitive(x + 1, y)) * scale;
+        yNormal = (getPrimitive(x, y) - getPrimitive(x, y + 1)) * scale;
         return new Vector3(xNormal, 1, yNormal).normalize();
     }
 
--- a/shared/src/main/java/com/faforever/neroxis/mask/NormalMask.java
+++ b/shared/src/main/java/com/faforever/neroxis/mask/NormalMask.java
@@ -35,8 +35,7 @@ public final class NormalMask extends VectorMask<Vector3, NormalMask> {
     }
 
     public NormalMask(FloatMask other, float scale, String name) {
-        this(other.getSize(), other.getNextSeed(), name, other.isParallel());
-        assertCompatibleMask(other);
+        this(other.getSize() - 1, other.getNextSeed(), name, other.isParallel());
         enqueue(dependencies -> {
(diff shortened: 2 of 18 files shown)


shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -206,20 +206,8 @@ private void setPrimitive(Vector2 location, float value) {
...
-        }
-        if (y == 0) {
-            yNormal = (getPrimitive(x, y) - getPrimitive(x, y + 1)) * scale;
-        } else if (y == (getSize() - 1)) {
-            yNormal = (getPrimitive(x, y - 1) - getPrimitive(x, y)) * scale;
-        } else {
-            yNormal = (getPrimitive(x, y - 1) - getPrimitive(x, y + 1)) * scale / 2f;
-        }
+        xNormal = (getPrimitive(x, y) - getPrimitive(x + 1, y)) * scale;
+        yNormal = (getPrimitive(x, y) - getPrimitive(x, y + 1)) * scale;
```
Sheikah45: Why did this function change? This change breaks the preservation of symmetry

shared/src/main/java/com/faforever/neroxis/mask/NormalMask.java
```diff
@@ -35,8 +35,7 @@ public NormalMask(FloatMask other) {
     }
 
     public NormalMask(FloatMask other, float scale, String name) {
-        this(other.getSize(), other.getNextSeed(), name, other.isParallel());
-        assertCompatibleMask(other);
+        this(other.getSize() - 1, other.getNextSeed(), name, other.isParallel());
```
Sheikah45, line 38: Why was the assertCompatible removed?

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
```diff
@@ -171,23 +171,32 @@ public static void writeRawDDS(BufferedImage image, Path path) throws IOExceptio
         Files.write(path, imageBytes.array(), StandardOpenOption.APPEND);
     }
 
-    public static BufferedImage getMapwideTexture(NormalMask normalMask, FloatMask waterDepth, FloatMask shadowMask) {
-        if (shadowMask.getSize() != normalMask.getSize()) {
-            throw new IllegalArgumentException("Mask sizes do not match: shadow size %d, normal size %d"
-                                                       .formatted(shadowMask.getSize(), normalMask.getSize()));
-        }
+    public static BufferedImage getMapInfoTexture(FloatMask waterDepth, FloatMask shadowMask) {
         waterDepth.resample(shadowMask.getSize());
```
Sheikah45: This should be using a copy of waterDepth not changing the size of the passed in mask if it needs to be rescaled just for this operation

Sheikah45 (changes requested):

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
BlackYps: The main purpose here is to change the normal calculation. It's a bit different to explain without a diagram, but I will try.
Before it calculated the normal at a point by using the difference between the pixel before the point and the pixel after the point. This results in a normal map with exactly the size of the heightmap input. It also makes the heightmap a bit blurry.
The accurate thing to do is to calculate the normal vector between two input points, then the normal vector is actually the normal vector of the plane between the points. This necessarily means that the resulting normal map is 1 smaller than the input heightmap. And the game actually uses a heightmap of n²+1 and a normal map of n².
So in essence this makes the calculated normal map more accurate, because it gets rid of the inherent smoothing of the old method.

The resulting normal map is symmetric if the underlying heightmap has been symmetric

shared/src/main/java/com/faforever/neroxis/mask/NormalMask.java
BlackYps: The two masks are not compatible anymore (in the classic sense) because the heightmap size is the size of the normalmap plus one.

shared/src/main/java/com/faforever/neroxis/util/ImageUtil.java
BlackYps: because it would also change the original floatmask even out of the scope of this function, right?

Sheikah45: Yes

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
Sheikah45: Unfortunately what you say about the symmetry is not true. 

If you have a 1x4 grid where the two inner rows are raised. So

0 1 1 0

The normal of this as above ends up being

1 0 -1 0 

This breaks the symmetry.

Sheikah45: Although I guess normals are not symmetric by nature. I have to look a little bit to determine the impact

BlackYps: The normal in your example would be 1 0 -1. It is symmetric because the normal texture has a smaller resolution. Although I noticed a different problem that I have to fix, where the functionn indeed produces results that are not symmetric.

Sheikah45: Yes but that is also a change where previously they were the same size.

Sheikah45: There aren't really other mask functions that have the side effects of returning a different size.

BlackYps: Do you take issue with the changed size or was that just an informational statement?

BlackYps: I prepared a quick graphic to explain the bugfix I just did:
grafik
Grey are the heightmap pixels and yellow is the normal map pixel. In my first commits I just calculated x1 and y1 and used them as the normal vector components. But in fact we have to calculate x1 and x2 and average them, same for y, to get a correct result where all neighboring heightmap pixels are taken into account.

Sheikah45: I think the changed size is non optimal as it is a bit of a gotcha going. But I guess the normals are also used in special circumstances in the first place. So I am not sure yet

BlackYps: yes, it's unusual, but it's just a fact of the game that the heightmap has a resolution of 2^n+1 while textures have a resolution of 2^n. Before this change we had to scale one of those two masks by one pixel which is awkward as well.

BlackYps:
Small reminder that the sunset biome needs to be disabled before the next game patch at the 16th of May. If you don't have the time to review the changes to the normal map calculations then I can make a separate pull request with just the commit that disables the biome.

Sheikah45 (approved):

PR #438, opened 2025-03-19
Multiple Octive Perlin Noise, Terrain Generator

clivepaterson:
Adds a new terrain generator, using multiple perlin noise layers added together, where the noise scale increases for each layer.

The theory:
Uses this concept: https://devforum.roblox.com/t/creating-procedural-mountains-a-fractal-noise-tutorial/1494362
And also domstrated here: https://github.com/PierfrancescoSoffritti/HeightMapGenerator/blob/master/README.md

So I made a terrain generator out of this stuff,
It uses up to 7 layers of Perlin noise...   (5K maps use 5 layers..)
image

And this is some previews for some maps that it generates:
5K multi_level
image
10K multi_level
image
20K multi_level
image

5K flooded_multi_level
image
10K flooded_multi_level
image
20K flooded_multi_level
image

20K flooded:
image

### DIFF
--- a/generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
@@ -4,7 +4,7 @@ public class RiversAndOceansMexResourceGenerator extends BasicResourceGenerator 
 
     @Override
     public void setupPipeline() {
-        resourceMask.init(passableLand).startVisualDebugger();
+        resourceMask.init(passableLand);
         resourceMask.add(passableWater
                                  .copy()
                                  .acid(random.nextFloat(0.3f, 0.7f), 0)
--- a/generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
@@ -4,8 +4,11 @@ public class RiversAndOceansMexResourceGenerator extends BasicResourceGenerator 
 
     @Override
     public void setupPipeline() {
-        resourceMask.init(passableLand);
-        resourceMask.add(passableWater);
+        resourceMask.init(passableLand).startVisualDebugger();
+        resourceMask.add(passableWater
+                                 .copy()
+                                 .acid(random.nextFloat(0.3f, 0.7f), 0)
+                                 .erode(random.nextFloat(0.5f,0.7f)));
         resourceMask.subtract(unbuildable);
         waterResourceMask.init(resourceMask);
     }
--- a/generator/src/main/java/com/faforever/neroxis/generator/terrain/FloodedMultiLevelTerrainGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/terrain/FloodedMultiLevelTerrainGenerator.java
@@ -1,0 +1,57 @@
+package com.faforever.neroxis.generator.terrain;
+
+import com.faforever.neroxis.generator.GeneratorParameters;
+import com.faforever.neroxis.map.SCMap;
+import com.faforever.neroxis.map.SymmetrySettings;
+import com.faforever.neroxis.mask.BooleanMask;
+import com.faforever.neroxis.util.vector.Vector2;
+import com.faforever.neroxis.util.vector.Vector3;
+
+public class FloodedMultiLevelTerrainGenerator extends MultiLevelTerrainGenerator {
+    @Override
+    public void initialize(SCMap map, long seed, GeneratorParameters generatorParameters,
+                           SymmetrySettings symmetrySettings) {
+        super.initialize(map, seed, generatorParameters, symmetrySettings);
+
+        waterHeight -= plateauHeight - 2f;
+        spawnLandHeight = landHeight + plateauHeight;
+
+        landNoiseMapFirstLevel = 0.40f;
+        landNoiseMapSecondLevel = 0.52f;
+        landNoiseMapThirdLevel = 0.61f;
+    }
+
+    @Override
+    protected void mountainSetup() {
+        mountains.setSize(map.getSize() + 1);
+    }
+
+    @Override
+    protected void landSetup() {
+        super.landSetup();
+        int mapSize = map.getSize();
+
+        if (mapSize < 512) {
+            BooleanMask connectionsMask = connections.copy().dilute(1, 18).setSize(land.getSize());
+            land.add(connectionsMask);
+            secondLevelLand.add(connectionsMask);
+        }
+
+        String[] SPAWN_MASK_BRUSHES = {
+                "mountain4.png",
+                "mountain6.png",
+                "mountain7.png",
+                "mountain8.png",
+                "mountain9.png",
+        };
+        map.getSpawns().forEach(spawn -> {
+            if (spawn.getTeamID() == 0) {
+                Vector3 location = spawn.getPosition();
+                String brush = SPAWN_MASK_BRUSHES[StrictMath.abs(random.nextInt()) % SPAWN_MASK_BRUSHES.length];
+                System.out.println("Adding " + brush + " to " + location.toString());
+                land.addBrush(new Vector2(location.x(), location.z()), brush, 15f, 256f, 200);
+                secondLevelLand.addBrush(new Vector2(location.x(), location.z()), brush, 15f, 256f, 200);
+            }
+        });
+    }
+}
--- a/generator/src/main/java/com/faforever/neroxis/generator/terrain/FloodedMultiLevelTerrainGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/terrain/FloodedMultiLevelTerrainGenerator.java
@@ -14,7 +14,6 @@ public class FloodedMultiLevelTerrainGenerator extends MultiLevelTerrainGenerato
         super.initialize(map, seed, generatorParameters, symmetrySettings);
 
         waterHeight -= plateauHeight - 2f;
-        spawnLandHeight = landHeight + plateauHeight;
 
         landNoiseMapFirstLevel = 0.40f;
         landNoiseMapSecondLevel = 0.52f;
@@ -29,26 +28,18 @@ public class FloodedMultiLevelTerrainGenerator extends MultiLevelTerrainGenerato
     @Override
     protected void landSetup() {
         super.landSetup();
-        int mapSize = map.getSize();
 
+        int mapSize = map.getSize();
         if (mapSize < 512) {
             BooleanMask connectionsMask = connections.copy().dilute(1, 18).setSize(land.getSize());
             land.add(connectionsMask);
             secondLevelLand.add(connectionsMask);
         }
 
-        String[] SPAWN_MASK_BRUSHES = {
-                "mountain4.png",
-                "mountain6.png",
-                "mountain7.png",
-                "mountain8.png",
-                "mountain9.png",
-        };
         map.getSpawns().forEach(spawn -> {
             if (spawn.getTeamID() == 0) {
                 Vector3 location = spawn.getPosition();
                 String brush = SPAWN_MASK_BRUSHES[StrictMath.abs(random.nextInt()) % SPAWN_MASK_BRUSHES.length];
-                System.out.println("Adding " + brush + " to " + location.toString());
                 land.addBrush(new Vector2(location.x(), location.z()), brush, 15f, 256f, 200);
                 secondLevelLand.addBrush(new Vector2(location.x(), location.z()), brush, 15f, 256f, 200);
             }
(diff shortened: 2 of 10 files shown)


clivepaterson:
So I figure I should post a map of the day. While I'm testing this (ie, generating a lot of maps, lol)
So this is a land one:
image

And the Water maps seem to work really well.
image
image

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -147,6 +148,10 @@ void setPrimitive(int x, int y, float value) {
     public FloatMask addPerlinNoise(int resolution, float scale) {
         int size = getSize();
         int gradientSize = size / resolution;
+        if (gradientSize <= 0) {
+            System.err.println("FloatMask:addPerlinNoise(): resolution "+resolution+" can't be greater than mask size " + size);
+            System.exit(2);
```
clivepaterson: I so will confess. That...
I know you won't like the System.exit() here...
And yes, I'm not a fan either.....   (I can remove this)

So anyway... I ran into this exception again... geeeees'..
It was because I used 7 octaves of perlin noise on a 5k map.... 

The output when you get this error looks like this:
image

Which is pretty hard output to interpret. I guess it's the mutlie threadedness which make the output a bit harder to read.
I wonder if Virtual Threads produce diffreent stacktrace's output, but Idk, I haven't used them before.

So anyway, with this System.exit() call here.... the error output looks like this:
image


I can remove this. It's basically just nice output for any future programmer that uses this function and run into the issue/constraint.

clivepaterson:
And might as well post a map(s) of the day.
This one is a huge 20k 6v6 map:
image
And another huge 20k 6v6 map:
image

image

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelTerrainGenerator.java
```diff
@@ -0,0 +1,221 @@
...
+    protected void landSetup() {
+        int mapSize = map.getSize();
+
+        int MAX_OCTAVES = 7;
+        int numOctaves = 0;
+        int smallestDetail = 5;
+
+        while (numOctaves < MAX_OCTAVES && smallestDetail << numOctaves <= mapSize) {
+            numOctaves++;
+        }
```
clivepaterson, line 64: This while loop finds the maximum number of octave's we can use for a given mapsize.
At 5k mapsize we have a maximum octave of 5, otherwise the Perlin noise function blows up.
And a MAX_OCTAVE of 7 seemed to be a good number for 20k maps.

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelTerrainGenerator.java
```diff
@@ -0,0 +1,221 @@
...
+
+        FloatMask landNoiseMap = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(), "landNoiseMap", true);
+        for (int octave = 0; octave < numOctaves; octave++) {
+            FloatMask octaveNoise = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(), "landNoiseOctave" + octave, true);
+            octaveNoise.addPerlinNoise(smallestDetail << octave,1f / numOctaves);
+            landNoiseMap.add(octaveNoise);
+        }
+        landNoiseMap.blur(8);
+
```
clivepaterson, line 73: And this give's us all the Perlin noise's added together:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelTerrainGenerator.java
```diff
@@ -0,0 +1,221 @@
...
+                .erode(0.3f, 10)
+                .setSize(mapSize+1);
+        secondLevelLand = landNoiseMap
+                .copyAsBooleanMask(secondLevel)
+                .erode(0.3f, 10)
+                .setSize(mapSize+1);
+        thirdLevelLand = landNoiseMap
+                .copyAsBooleanMask(thirdLevel)
+                .erode(0.3f, 10)
+                .setSize(mapSize+1);
```
clivepaterson, line 89: And then we take 3 slices of the noise map at slightly random heights to give variation the 2nd, and 3rd layers.
And we get something like this:
image
image
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelTerrainGenerator.java
```diff
@@ -0,0 +1,221 @@
...
+
+    @Override
+    protected void plateausSetup() {
+        int mapSize = map.getSize();
+        plateaus.setSize(mapSize + 1);
+    }
+
+    @Override
+    protected void initRamps() {
+        ramps.setSize(map.getSize() + 1);
```
clivepaterson, line 123: This function uses the Outline of the 2nd layer of Land:
image

And then uses and Perlin noise to decide what edges of the 2nd level should connect to down the 1st level using ramps:

Perlin Noise Mask: (The nice thing about using perlin noise here is that it kinda garantee's connections)
image

And the Ramp Mask comes out like this:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelTerrainGenerator.java
```diff
@@ -0,0 +1,221 @@
...
+        FloatMask rampExclusion = new FloatMask(ramps.getSize(), random.nextLong(), this.symmetrySettings, "rampExclusion", true);
+        rampExclusion.addPerlinNoise(StrictMath.min(128, rampExclusion.getSize()), 1);
+        BooleanMask rampExclusionMask = rampExclusion.copyAsBooleanMask(0.4f, 0.6f);
+
+        ramps.subtract(rampExclusionMask.invert());
+    }
+
+    @Override
+    protected void blurRamps() {
+        BooleanMask inflatedRamps = ramps.copy().subtract(spawnLandMask);
```
clivepaterson, line 137: And I had to override this function and adjust things, it's a tricky function to get right.
But the final results are this:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelTerrainGenerator.java
```diff
@@ -0,0 +1,221 @@
...
+                 .blur(32, inflatedRamps.inflate(2))
+                 .blur(4, inflatedRamps.copy().outline().inflate(4))
+                 .blur(6, inflatedRamps.copy().outline().inflate(6))
+                 .clampMin(0f)
+                 .clampMax(255f);
+    }
+
+    @Override
+    protected void setupHeightmapPipeline() {
+        int mapSize = map.getSize();
```
clivepaterson, line 148: So this is very similar to the base function.
The only thing that changes is that this line gets removed:
image
And we add this:
image

This allows players to start on a plateua.

Without this I was finding some spawns the were completely surrounded by walls.
Like this but worse:
image

But with this change, we get a evalvated spawn location:
image

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -419,6 +424,38 @@ public FloatMask useBrushWithinArea(BooleanMask other, String brushName, int siz
...
     }
 
+    /**
+     * Take the height value for each spawn point provided, and set the surrounding terrain to the same height, within
+     * a brush mask.
+     * @param spawns    The list of spawns
+     * @param brush     The brush to use for each spawn, this brush masks off the surrounding terrain to be leveled
+     * @param radius    The radius of the spawn mask to use for flattening
+     */
+    public FloatMask flattenSpawnPointsWithRadius(List<Spawn> spawns, String brush, int radius) {
```
clivepaterson: So this is the flatten function.
It took a couple goes before I used the Symmetry which is the more correct implementation.
Anyway, this was a tricky function to implement but the result are good.
image

generator/src/main/java/com/faforever/neroxis/generator/resource/RiversAndOceansMexResourceGenerator.java
```diff
@@ -5,7 +5,10 @@ public class RiversAndOceansMexResourceGenerator extends BasicResourceGenerator
     @Override
     public void setupPipeline() {
         resourceMask.init(passableLand);
-        resourceMask.add(passableWater);
+        resourceMask.add(passableWater
+                                 .copy()
+                                 .acid(random.nextFloat(0.3f, 0.7f), 0)
+                                 .erode(random.nextFloat(0.5f,0.7f)));
```
clivepaterson, line 11: This change reduces the amount of mexes in the water. I was finding way too many mexes in the water...
So this uses the .acid().erode() trick to reduce the likelyhood of water mexes:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/FloodedMultiLevelTerrainGenerator.java
```diff
@@ -0,0 +1,48 @@
...
+        landNoiseMapThirdLevel = 0.61f;
+    }
+
+    @Override
+    protected void mountainSetup() {
+        mountains.setSize(map.getSize() + 1);
+    }
+
+    @Override
+    protected void landSetup() {
```
clivepaterson, line 29: So this flooded Mutli level, need to garantee that all spawn locations start on Level 2, because level 1 gets flooded by the water height.

And I also added Land Connections for smaller maps like 5K maps to make them more playable.

clivepaterson (commented):
The Epic self review, I think the code is bascially ready now.
I'm also curioous what @BlackYps feels.

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
Sheikah45: I understand why you would want to use the system.exit but unfortunately we can't since it is possible this is called inside another jvm instead of just as a standalone. Really this should be fixed by using actual logging instead of just usint System,.out and System.err

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
Sheikah45: I like this. Probably have to investigate using this more generally instead of the forcing of the map we do now. Only thing I will say is that this should probably go in MapMaskMethods instead of FloatMask directly as MapMaskMethods is where methods that interact with map properties go.

Sheikah45 (changes requested):

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelTerrainGenerator.java
clivepaterson: Oh yep, there's also this bit:
image

Which is where we add the 1st, 2nd, 3rd layers of land... so yeah I guess this is the main difference for this function.

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
clivepaterson: Hmmm, very very good point. lol...
So I did a take on on this... and thrown a RuntimeException instead.

I noticed that there's a couple functions below `getMin()` and `getMax()` which throw an IllegalStateException which inherits RuntimeException...

So maybe a runtime exception is OK?

Othwise System.out seems to be the main thing used in the codebase, was looking around where it might use a logger of some sort, but I think that would have to be added.....

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
clivepaterson: Ok I see.
I have moved this function into MapMaskMethods now.
And tested.... all good.

This was a good one from testing:
image

Most spawns are still on the bottom layer.

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
clivepaterson: And the output when I make this expcetion occur looks like this:
image

Sheikah45: Yeah it doesn't use a logger yet that has been on my to-do list for a while....

clivepaterson: Well I guess System.err for now... ;-) lol.
But I guess it documents the condiction in the code at least.

clivepaterson:
Well, some last minute small changes today.
But testing is looking all good. I feel this must be close to merging...

It's quite a different style / vibe with these maps. Will be interesting to see how the players react.

And so, with that, I present today's map of the day...
This one is a 7.5k 1v1.
image

BlackYps:
I don't really have time to look at it in detail, but I like the addition of fractal noise. It's a powerful basic feature to have for terrain generation.

Sheikah45 (approved):

PR #439, opened 2025-04-02
Some fixes: 

clivepaterson:
1) I forgot to put Multilevel terrain generators in the TerrainStyle enum
   Missing in the list oops... no biggy, it's still there in the Map Style list
   
image

2) Improved spawn blur in flattenSpawnPointsWithRadius()
The blur was basically not smooth enough... needs more smooother.
Before:
Spawn_blur_unsmooth
After:
Spawn_blur_smoothed_more

3) Improved the ramp blur for Multilevel terrain generator
This one is really subtle but still bugs me when I see it.
Before:
image
After:
image

4) fixed the mex placement too close to water for RiversAndOceans mex generator.
Ok this is kinda actually bad, I was watch some Live Replay on release night, watching the laddre players try out these new maps:
And I saw this one:
image

So mexes were allowed to generate too close to the water.... not good.
Specially since 1 team had 1 less mex...
I don't even know how I managed to make this happen...
Anyway the fix for this is in RiversAndOceansMexResourceGenerator.java  (more space between ocean and mexes)
The same seed looks like this now:   (Very different layout, I guess this is the butterfly effect in action)
image


clivepaterson:
Bump @Sheikah45

Issue #440, opened 2025-04-03
Rivers terrain has too many small rivers for good gameplay


lL1l1:
Particularly on large maps and with less spawns, Rivers terrain generates too many rivers, which results in:
- Dominance by hover units since they're the only ones that can fight across frequent, small water obstacles. Cybran has no hover so it is especially hurt by this. UEF and Cybran have no hover flak so T2 gunships are too powerful against them.
  - Also, amphibious unit (which move underwater) balance is a bit skewed by the terrain style in favor of shorter range amphibious units.
  - Overall, it doesn't seem like such heavy emphasis on hover should be encouraged.
- There seem to be few mexes that can be contested because of water occupying mex spaces and making natural chokepoints for attacking units.

I think the solution to these issues would be to have fewer rivers that are longer and just slightly wider to avoid tiny ponds. They should also curve or split off less so that they don't form islands as much where you have to pass over water multiple times in a short distance.

These are my thoughts after some [comments others made on Discord](https://discord.com/channels/197033481883222026/1089784297789083730/1356822914414018651) when I brought up the new terrain style.

## Some Examples:

1. This is 10km 1v1 (few players for its size) and it has extremely disruptive and chaotic terrain. The mexes are also very easy to defend with large distance inbetween and being clumped,

`neroxis_map_generator_1.14.0_xfzfbxctmhydk_aieaeaqabmbaavts`
![Image](https://github.com/user-attachments/assets/05e13144-ac91-4efc-a658-1e105d6d2e60)

2. 10km 2v2: there are few mexes to fight over, and most of the middle is dominated by impassable terrain.

`neroxis_map_generator_1.14.0_guunk3474sjmg_aqeaeeyebmbaislx`
![Image](https://github.com/user-attachments/assets/fe330f3a-2604-473c-a2f7-1fa09b2c06e4)

3. 20km 8v8: land bridges are very small and few between, along with many water gaps. There are even ponds in the player bases which is a bit annoying with how small they are, which looks purposeless.

`neroxis_map_generator_1.14.0_jjwbqmir7wlgy_caiaeaaibmbae4lx`
![Image](https://github.com/user-attachments/assets/5fd3b2d1-bacb-424c-85a1-2ee10ab18b38)

4. 16.5km 4v4: Hover seems to be extremely dominant due to large amount of tiny branches the rivers spawn or the small strips of water around the center of the map.

[...]

PR #441, opened 2025-04-15
Changes to rivers terrain gen

clivepaterson:
Ok so new day, new fixes...
This one comes from criticism of the rivers terrain generator, (the origin of all the shinanigans)
So ultimately the Rivers can produce some bad maps... (not always, but sometimes)
And there some issues as discussed in https://github.com/FAForever/Neroxis-Map-Generator/issues/440

So I've reduced it's prevalance on ladder...

And we've tried to fix the main problem.
Anyway, as always, I gotta spam some pitcures....
This is the next 4 random maps I generated with the changes:
image
Almost like Twin rivers this one: (no not really)
image

image
image

So I hope this addresses the issues.. I should definitely pr these fixes.

Maybe I can provide some bfore and after screen shots if needed.

### DIFF
--- a/generator/src/main/java/com/faforever/neroxis/generator/MapStyle.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/MapStyle.java
@@ -37,12 +37,12 @@ public enum MapStyle {
     LOW_MEX(LowMexStyleGenerator.class, LowMexStyleGenerator::new, .5f),
     MOUNTAIN_RANGE(MountainRangeStyleGenerator.class, MountainRangeStyleGenerator::new, 1),
     MULTILEVEL(MultiLevelStyleGenerator.class, MultiLevelStyleGenerator::new, 1),
-    FLOODED_MULTILEVEL(FloodedMultiLevelStyleGenerator.class, FloodedMultiLevelStyleGenerator::new, 1),
+    FLOODED_MULTILEVEL(FloodedMultiLevelStyleGenerator.class, FloodedMultiLevelStyleGenerator::new, 0.75f),
     ONE_ISLAND(OneIslandStyleGenerator.class, OneIslandStyleGenerator::new, 1),
     SMALL_ISLANDS(SmallIslandsStyleGenerator.class, SmallIslandsStyleGenerator::new, 4),
     VALLEY(ValleyStyleGenerator.class, ValleyStyleGenerator::new, 1),
-    RIVERS(RiversStyleGenerator.class, RiversStyleGenerator::new, 1),
-    RIVERS_AND_OCEANS(RiversAndOceansStyleGenerator.class, RiversAndOceansStyleGenerator::new, 1);
+    RIVERS(RiversStyleGenerator.class, RiversStyleGenerator::new, 0.1f),
+    RIVERS_AND_OCEANS(RiversAndOceansStyleGenerator.class, RiversAndOceansStyleGenerator::new, 0.75f);
 
     private final Class<? extends StyleGenerator> generatorClass;
     private final Supplier<StyleGenerator> generatorSupplier;
--- a/generator/src/main/java/com/faforever/neroxis/generator/MapStyle.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/MapStyle.java
@@ -27,7 +27,7 @@ import java.util.function.Supplier;
 @AllArgsConstructor
 public enum MapStyle {
     BASIC(BasicStyleGenerator.class, BasicStyleGenerator::new, 1),
-    BIG_ISLANDS(BigIslandsStyleGenerator.class, BigIslandsStyleGenerator::new, 4),
+    BIG_ISLANDS(BigIslandsStyleGenerator.class, BigIslandsStyleGenerator::new, 1),
     CENTER_LAKE(CenterLakeStyleGenerator.class, CenterLakeStyleGenerator::new, 1),
     DROP_PLATEAU(DropPlateauStyleGenerator.class, DropPlateauStyleGenerator::new, .5f),
     FLOODED(FloodedStyleGenerator.class, FloodedStyleGenerator::new, .01f),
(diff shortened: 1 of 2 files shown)


generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -28,7 +28,9 @@ public void initialize(SCMap map, long seed, GeneratorParameters generatorParame
 
         mountainBrushSize = 24;
         mountainBrushDensity = 8f;
-        mountainBrushIntensity = 0.4f;
+        mountainBrushIntensity = 0.8f;
+
+        spawnSize = 48;
```
clivepaterson, line 33: So needed to fix the `mountainBrushIntensity` because units can sometimes walk over the mountains.. but a higher intensity here makes the mounatins taller and impassable.

And set the spawnSize which is used later in the `spawnMaskSetup()` which is what other terrain generators do.
And the spawnSize of 48 avoids having maintains/plateua's too close to a spawn.

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -42,17 +44,10 @@ protected void landSetup() {
...
         riverMask = rivers.copyAsBooleanMask(0.5f, 0.65f);
 
-        FloatMask riverExclusion = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(), "riversExclusion", true);
-        riverExclusion.addPerlinNoise(StrictMath.min(256, mapSize), 1);
-        float exclusionThickness = 0.06f * ((float)mapSize / 256f);
-        riverExclusionMask = riverExclusion.copyAsBooleanMask(0.5f, 0.5f + exclusionThickness);
-        riverMask.subtract(riverExclusionMask);
-        riverExclusionMask.setSize(mapSize+1);
-
         riverMask.invert();
```
clivepaterson, line 47: So turns out this exclusion mask doesn't really work well.
Not a bad idea, but the `connections` is a better way to make the map more playable instead of being a floaty only game.

So maybe in the end, this exclusion mask was just ruining the maps.

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -42,17 +44,10 @@ protected void landSetup() {
...
         riverMask.blur(10);
 
-        riverMask.add(connections.copy().dilute(1, 10).setSize(riverMask.getSize()));
+        riverMask.add(connections.copy().dilute(1, 20).setSize(riverMask.getSize()));
```
clivepaterson, line 50: This setting makes a big difference to the land units.
I'm also wondering if 25 might be a better value. It basically balancing the land vs floaty for the map.

Bascially, when the connection's thickness is bigger, then the map become more a more direct land assalt.
I think 20 is good. maybe a range of 20-25 might be better.

generator/src/main/java/com/faforever/neroxis/generator/terrain/RiversTerrainGenerator.java
```diff
@@ -219,7 +214,7 @@ protected void spawnTerrainSetup() {
     protected void spawnMaskSetup() {
         map.getSpawns().forEach(spawn -> {
             Vector3 location = spawn.getPosition();
-            spawnLandMask.fillCircle(location, 10, true);
+            spawnLandMask.fillCircle(location, spawnSize, true);
```
clivepaterson, line 217: Yeah, dunno why this was hard coded... 
Just left from some testing tweeeking playing maybe...  prolly been there for a long time...

generator/src/main/java/com/faforever/neroxis/generator/MapStyle.java
```diff
@@ -36,13 +36,13 @@ public enum MapStyle {
...
+    MULTILEVEL(MultiLevelStyleGenerator.class, MultiLevelStyleGenerator::new, 1f),
+    FLOODED_MULTILEVEL(FloodedMultiLevelStyleGenerator.class, FloodedMultiLevelStyleGenerator::new, 0.75f),
     ONE_ISLAND(OneIslandStyleGenerator.class, OneIslandStyleGenerator::new, 1),
-    SMALL_ISLANDS(SmallIslandsStyleGenerator.class, SmallIslandsStyleGenerator::new, 4),
+    SMALL_ISLANDS(SmallIslandsStyleGenerator.class, SmallIslandsStyleGenerator::new, 1),
     VALLEY(ValleyStyleGenerator.class, ValleyStyleGenerator::new, 1),
-    RIVERS(RiversStyleGenerator.class, RiversStyleGenerator::new, 1),
-    RIVERS_AND_OCEANS(RiversAndOceansStyleGenerator.class, RiversAndOceansStyleGenerator::new, 1);
+    RIVERS(RiversStyleGenerator.class, RiversStyleGenerator::new, 0.25f),
+    RIVERS_AND_OCEANS(RiversAndOceansStyleGenerator.class, RiversAndOceansStyleGenerator::new, 0.75f);
```
clivepaterson, line 46: Ok, I've rebalanced this... I belieive this affect the prevalance of mapstyle on Ladder.
@Sheikah45  Feel free to use admin and override any of these values..

But for me this feels like a better balance of maps for ladder.

clivepaterson (commented):
Self review

lL1l1:
I like the maps in the new screenshots, it seems like using connections over the exclusion mask was the right move.

The mountains are able to encompass an entire river or most of it, would it be possible to limit the mountains to be around a single point on the river, as if the water is coming out of the mountain at that location?

clivepaterson:
> I like the maps in the new screenshots, it seems like using connections over the exclusion mask was the right move.
> 
> The mountains are able to encompass an entire river or most of it, would it be possible to limit the mountains to be around a single point on the river, as if the water is coming out of the mountain at that location?

Hmm so the process for mountains is this..
If we take the following map:
neroxis_map_generator_snapshot_oonnihy7jphz2_ayfaefa_preview
Then we first create an outline of the rivers:
image
It then creates a mountain exclusion mask: (hmm this is very white, too white actually)
image
Anyway, after substracting the exclusion mask it leaves this:
image



As I'm writing this response, I'm not liking that exclusion mask.. So I tweaked it to look like this:
image
Which gives these mountains:
image
And the maps looks like this:
image

as for "water is coming out of the mountain", I'm not sure I could achieve this, lol. I have limited abilities, but I change change the mountain exlcusion mask which kinda controls how much a river is surrounded by mountains.

clivepaterson:
Bump @lL1l1 @Sheikah45 
Any thought's, feelings, suggestions?

I can make any tweaks you think are needed.... 

The weekend has started for me, so I'm prolly digging into some mapgen the next few days..
Like this one with 2 screen shots:
image
image
Still some bugs to iron out, lol. But I think it's more promising than the old Rivers one.... (Which is still OK I think).

Sheikah45:
Sorry I missed the original notification of the PR going up.

I think these are good changes. It is always hard to get that proper control over the floaty and real land connections.

I think the exclusion mask was a good attempt but if it was too sparse then yeah connections is the way to go.

It is all really an iterative process.

Sheikah45 (approved):

PR #455, opened 2025-07-20
Lots of adjustments to Multilevel Terrain, and added HeatMapResource generator

clivepaterson:


generator/src/main/java/com/faforever/neroxis/generator/prop/HeatMapPropGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+import com.faforever.neroxis.generator.GeneratorParameters;
+import com.faforever.neroxis.generator.terrain.TerrainGenerator;
+import com.faforever.neroxis.map.SCMap;
+import com.faforever.neroxis.map.SymmetrySettings;
+import com.faforever.neroxis.mask.BooleanMask;
+import com.faforever.neroxis.mask.FloatMask;
+import com.faforever.neroxis.util.DebugUtil;
+import com.faforever.neroxis.util.Pipeline;
+
+public class HeatMapPropGenerator extends BasicPropGenerator {
```
clivepaterson, line 12: This is  a new resource generator, which can use a FloatMask to generate tree groups and rock/boulders.
This is an example of a Heatmap created by a Multilayered terrain generator:
image
For the rocks I had to make a mask otherwise there was too much rocks. So I made a mask that looks like this:
image
Which creates these reclaim area's for rocks like this:
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelOpenTerrainGenerator.java
```diff
@@ -0,0 +1,46 @@
+package com.faforever.neroxis.generator.terrain;
+
+import com.faforever.neroxis.generator.GeneratorParameters;
+import com.faforever.neroxis.map.SCMap;
+import com.faforever.neroxis.map.SymmetrySettings;
+import com.faforever.neroxis.util.Pipeline;
+
+public class MultiLevelOpenTerrainGenerator extends MultiLevelTerrainGenerator {
```
clivepaterson, line 8: So this is a new Multilevel Terrain Generator based on my thoughts about the Ladder experience with these maps.

So this aims to provide a mostly Land battle map, with less terrain features.
I provides simple open maps leading to some open land battles.

The tradditional Multi Terrain Gnerator is designed to provide maps like Syrtis Major with 50% of the map being plateu.
This is mean't to provide an Open Land battle where 70-80% of the map is flat land.

Here are some screen shots of the maps generatored by this:

10K 3v3
neroxis_map_generator_snapshot_fqqlnfgpyn4r2_ayeaeey_preview
image

10K 3v3
neroxis_map_generator_snapshot_udv3ozeogp6es_ayeaeaa_preview
image

20K 5v5
neroxis_map_generator_snapshot_ealdpgjdwnqsm_biiaeey_preview
image

20K 5v5
neroxis_map_generator_snapshot_yr6y5cnp7lwco_biiaedy_preview
image

shared/src/main/resources/custom_biome/Crystalline/decals.json
```diff
@@ -12,7 +12,7 @@
     "/env/geothermal/decals/gt_noise02_normals.dds"
   ],
   "fieldAlbedos": [
-    "/env/crystalline/decals/scx1_proto01_crev02b_albedo.dds",
+    "/env/common/decals/tarmacs/tar8x_seraphim_01_albedo.dds",
```
clivepaterson, line 15: Replaces this decal:
image


With This one:
image
Looks better.

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelTerrainGenerator.java
```diff
@@ -62,40 +70,44 @@ public void initialize(SCMap map, long seed, GeneratorParameters generatorParame
...
-        int smallestDetail = 5;
 
-        while (numOctaves < MAX_OCTAVES && smallestDetail << numOctaves <= mapSize) {
+        while (numOctaves < MAX_OCTAVES && noiseSmallestDetail << numOctaves <= mapSize) {
             numOctaves++;
         }
 
-        FloatMask landNoiseMap = new FloatMask(mapSize, getRandom().nextLong(), land.getSymmetrySettings(),
-                                               "landNoiseMap", pipeline);
+        float amplitude = 1f;
```
clivepaterson, line 78: This amplitude is a new thing. A value of 1 mean's the algoithm is unchanged.

But I've been exploring the idea of setting the land height to the noise map.
Which looks like this when amplitude is 1f:
image

And this is what it looks like with 2f:
image

generator/src/main/java/com/faforever/neroxis/generator/prop/HeatMapPropGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+
+    protected FloatMask resourceDensityMap;
+
+    @Override
+    public void initialize(SCMap map, long seed, GeneratorParameters generatorParameters,
+                           SymmetrySettings symmetrySettings, TerrainGenerator terrainGenerator, Pipeline pipeline) {
+        super.initialize(map, seed, generatorParameters, symmetrySettings, terrainGenerator, pipeline);
+        pipeline.setDebug(true);
```
Sheikah45: This debug should be removed

generator/src/main/java/com/faforever/neroxis/generator/prop/HeatMapPropGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+        resourceDensityMap = terrainGenerator.getResourceDensityMap();
```
Sheikah45, line 20: I am not totally convinced that expanding thee public api of the terrain generator is the right move here. It doesn't quite make sense to me that the terrain generator is creating a resourceDensityMap, especially when most of the terrain generators don't produce one.

generator/src/main/java/com/faforever/neroxis/generator/prop/HeatMapPropGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+                                  .setValue(heatExlcusion, 0)
+                                  .blur(25)
+                                  .setValue(heatExlcusion, 0)
+                                  .blur(30);
+                resourceDensityMap.scaleToNewMinAndMaxHeight(0, 1);
+            }
+
+            resourceDensityMap.setValue(noProps.copy().inflate(4), 0);
+            Biome biome = map.getBiome();
+            propPlacer.placeProps(resourceDensityMap, biome.propMaterials(), reclaimDensity);
```
Sheikah45: Like this section confuses me. We don't really want to be creating masks in the placeProps function since it is synchronous. Additionally at this point the passableLand Pipeline may have finished or be in process of finishing so it is in a weird situation that could lead to determinism issues as well

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelOpenTerrainGenerator.java
Sheikah45: I am not sure I totally see the value in this new generator if it's focus is on mainly creating open land maps. I think a common criticism of the current generators in total is that they are too prone to just creating open land maps, so adding another generator like that might not be the best idea.

shared/src/main/java/com/faforever/neroxis/map/placement/PropPlacer.java
```diff
@@ -43,4 +48,83 @@ public void placeProps(BooleanMask spawnMask, List<String> paths, float minSepar
...
+
+    public void placeProps(FloatMask heatMap, PropMaterials propMaterials, float reclaimDensity) {
+        FloatMask rockAreas = new FloatMask(heatMap.getSize(), random.nextLong(), heatMap.getSymmetrySettings(), "Rock Areas", null);
+        rockAreas.addPerlinNoise(64, 1f);
+        BooleanMask rockAreaMask = rockAreas.copyAsBooleanMask(0.7f);
+
+        heatMap.loopInSymmetryRegion(SymmetryType.SPAWN, (x, y) -> {
+            float heat = heatMap.get(x, y);
+            float xGitter = x + random.nextFloat(1f) - 0.5f;
+            float yGitter = y + random.nextFloat(1f) - 0.5f;
```
Sheikah45: Gitter -> Jitter

shared/src/main/java/com/faforever/neroxis/map/placement/PropPlacer.java
```diff
@@ -43,4 +48,83 @@ public void placeProps(BooleanMask spawnMask, List<String> paths, float minSepar
...
+            Vector2 location = new Vector2(xGitter, yGitter);
+            ArrayList<Vector2> origAndSymmetryPoints = new ArrayList<>(
+                    heatMap.getSymmetryPoints(location, SymmetryType.SPAWN)
+                           .stream()
+                           .map(Vector::roundToQuarterPoint)
```
Sheikah45: I don't understand the purpose of rounding to the quarter point

shared/src/main/java/com/faforever/neroxis/map/placement/PropPlacer.java
```diff
@@ -43,4 +48,83 @@ public void placeProps(BooleanMask spawnMask, List<String> paths, float minSepar
...
+        int mapSize = map.getSize();
+        final int PADDING = 10;
+
+        locations.forEach(location -> {
+            if (location.x() < PADDING || location.x() > mapSize-PADDING || location.y() < PADDING || location.y() > mapSize-PADDING) {
+                withinBounds.set(false);
+            }
+        });
+
+        return withinBounds.get();
```
Sheikah45: You don't need an atomicBoolean here you can just use a stream.noneMatch and that also supports the short-circuiting we would want.

shared/src/main/java/com/faforever/neroxis/map/placement/PropPlacer.java
```diff
@@ -43,4 +48,83 @@ public void placeProps(BooleanMask spawnMask, List<String> paths, float minSepar
...
+
+                if (areAllPointsWithinPropBounds(origAndSymmetryPoints)) {
+                    origAndSymmetryPoints.forEach((point) -> {
+                        Prop prop = new Prop(finalPropPath, new Vector2(point.x(), point.y()), finalPropRotation, finalPropIsBoulder);
+                        map.addProp(prop);
+                    });
+                }
+            }
+        });
+    }
```
Sheikah45: I am not a big fan of exposing the loopInSymmetryRegion here as public API because it has considerations for pipelined masks and synchronization.

It would probably be better to add some kind of peek method that can work more generally.

shared/src/main/java/com/faforever/neroxis/map/placement/PropPlacer.java
```diff
@@ -43,4 +48,83 @@ public void placeProps(BooleanMask spawnMask, List<String> paths, float minSepar
...
+            }
+
+            if (propPath != null) {
+                String finalPropPath = propPath;
+                float finalPropRotation = propRotation;
+                boolean finalPropIsBoulder = propIsBoulder;
+
+                if (areAllPointsWithinPropBounds(origAndSymmetryPoints)) {
+                    origAndSymmetryPoints.forEach((point) -> {
+                        Prop prop = new Prop(finalPropPath, new Vector2(point.x(), point.y()), finalPropRotation, finalPropIsBoulder);
```
Sheikah45: Vectors are immutable so you don't have to create a new one

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -626,6 +626,17 @@ protected void set(int x, int y, Float value) {
...
+    public FloatMask setValue(BooleanMask areaToSet, float value) {
+        return enqueue(dependencies -> {
+            BooleanMask maskForSetting = (BooleanMask) dependencies.getFirst();
+            apply((x, y) -> {
+                if (maskForSetting.get(x,y)) {
+                    mask[x][y] = value;
+                }
+            });
+        }, areaToSet);
+    }
```
Sheikah45: This function already exists. It is SetToValue, and if it is for the primitive then it should be named setToPrimitive and also use the setPrimitive and getPrimitive functions. Interactions with the mask should go through those functions rather than direct access

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -665,6 +676,37 @@ protected FloatMask setSizeInternal(int newSize) {
...
+            float[][] oldMask = mask;
+            initializeMask(getSize());
+            applyWithSymmetry(SymmetryType.SPAWN, (x, y) -> {
+                float oldValue = oldMask[x][y];
+                float newValue = (oldValue - oldMin) * scale + newMin;
+                applyAtSymmetryPoints(x, y, SymmetryType.SPAWN, (sx, sy) -> {
+                    set(sx, sy, newValue);
+                });
+            });
+        });
```
Sheikah45: This method doesn't need to create a copy of the oldMask to reference the new mask. It is a function of purely the existing value at a pixel. So it should just use the set method. Additionally it doesn't need to use the with symmetry variant since it maintains symmetry naturally by being a 1 to 1 pixel function

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -665,6 +676,37 @@ protected FloatMask setSizeInternal(int newSize) {
...
+            float[][] oldMask = mask;
+            initializeMask(getSize());
+            applyWithSymmetry(SymmetryType.SPAWN, (x, y) -> {
+                float oldValue = oldMask[x][y];
+                float newValue = (float)StrictMath.pow(oldValue, exp);
+                applyAtSymmetryPoints(x, y, SymmetryType.SPAWN, (sx, sy) -> {
+                    set(sx, sy, newValue);
+                });
+            });
+        });
```
Sheikah45: This method doesn't need to create a copy of the oldMask to reference the new mask. It is a function of purely the existing value at a pixel. So it should just use the set method. Additionally it doesn't need to use the with symmetry variant since it maintains symmetry naturally by being a 1 to 1 pixel function

shared/src/main/java/com/faforever/neroxis/mask/MapMaskMethods.java
```diff
@@ -153,4 +153,38 @@ public static FloatMask flattenPointsWithRadius(List<Vector2> locations, FloatMa
...
+                float value = noise.get(x, y);
+
+                if (value > minHeight && value < maxHeight) {
+                    // Add a value between 0 and 1 for the heat map, anything outside of min and max height will be 0.
+                    return existingValue + 1 - (StrictMath.abs(value - middleValue  ) / range);
+                } else {
+                    return existingValue;
+                }
+            });
+        }, noiseMap);
```
Sheikah45, line 174: I don't know if this function is general enough to be in MapMaskMethods. It can probably just be a function in the MultiLevelTerrainGenerator

shared/src/main/java/com/faforever/neroxis/mask/MapMaskMethods.java
```diff
@@ -153,4 +153,38 @@ public static FloatMask flattenPointsWithRadius(List<Vector2> locations, FloatMa
...
+           exec.setPrimitiveWithSymmetry(SymmetryType.SPAWN, (x, y) -> {
+               if (flattenMask.get(x, y)) {
+                   return destinationHeight;
+               } else {
+                   return exec.getPrimitive(x, y);
+               }
+           });
+           exec.blur(1, flattenMask.outline());
+        });
+    }
```
Sheikah45: I don't think this function is necessary because it can be easily done with
BooleanMask flattenMask = exec.setToValue(exec.copyAsBooleanMask(minHeight, maxHeight);
exec.setToValue(flattenMask, destinationHeight).blur(1, flattenMask.outline()).

The above also allows for the operations to be parallelized as possible as well.

Sheikah45 (changes requested):

generator/src/main/java/com/faforever/neroxis/generator/prop/HeatMapPropGenerator.java
clivepaterson: Yep is is only used by the the noise maps.

The concept was to use some of the noise data to produce the areas of tree groups and rocks etc.
The main difference from other PropPlacers's is that:
- most of the rocks appear away from mountain edges, ie more between the mountain edges water ponds.
- There's a lot of tree groups, (a resource density of about 0.7 is the same amount of trees on a 20K as Setons) (I've increase the amount of tree groups now... should be better)

I did create a nice fallback HeatMap for other terrain generator's. It used the resourceMask and some blur()'s to create the same affect. But this can only by done by via custom maps settings.

But it's a fairly good prop placer.
It's definitely different.
It's not next level amazing, but it's different, and it's not bad I think.

This is a screen shot of a 20K map with the HeatMap: (It doesn't show the tree groups, but they are there)
image

generator/src/main/java/com/faforever/neroxis/generator/prop/HeatMapPropGenerator.java
clivepaterson: Yep, this was reallly just a fallback option, incase HeatMap is selected in FAF when making a custom map. (User can still select this option)

I have changed it to just use `super.setupPipeline();` and `super.placeProps();` incase someone selects this option.
I hope using `super.` is ok for the pipeline/determinism
But yeah, I didn't realise it would mess up the pipeline, my bad.... it's complicated... lo.l.

clivepaterson:
So converting to draft and pushing on with the fractal genertor.
Which is still a work in progress.
But this is the maps that it's generating at the moment:
image

And here is a couple more maps:
neroxis_map_generator_snapshot_t3q7yaj2pdakw_biiaeey_preview
image

neroxis_map_generator_snapshot_36vcdb3gkvb2m_biiaeaa_preview
image

neroxis_map_generator_snapshot_6djazz22ac4zc_biiaeea_preview
image

generator/src/main/java/com/faforever/neroxis/generator/prop/HeatMapPropGenerator.java
Sheikah45: Can we generate the resourceDensityMask in the PropGenerator?

PR #457, opened 2025-07-30
Improve default values for PBR texture generation

BlackYps:
I want to improve the experience when you use the neroxis toolsuite to generate the PBR texture and you don't provide textures for all layers. This can easily be the case when creating a map, when you want to do first tests with only a few textures.
A 0.5 value for heightmaps results in the texture layer behaving almost like the traditional linear interpolation blending. Previously, the 0 value would make the texture layer vanish.
A 0.5 value for roughness results in a surface with noticable reflections, without appearing extremely shiny.
Previously, the 0 value would make the surface look extremely shiny as if it was dripping wet.


BlackYps:
I tested it and it works correctly now

Sheikah45 (approved):

PR #462, opened 2025-08-29
Fractal generator rebased

clivepaterson:
Some previews of the fractal generantor:
image
image

And an example from loaded in the Map Editor:
image

And an example of the Blur along the symmetry line: (This seems to do a good enough job of hiding the symmetry artifacting)
image

I made a bunch of screen shots today, to show how the fractal generator builds the terrain step by step:
image


clivepaterson:
I have converted this PR to a draft, again.
The massive cliff faces that get produced for rotational symmetries is pretty bad:
Eg:
image

But I have a method of detecting these cliff edges:
image

And so I'm going to use the Mountain generation pipline to add mountains over these cliff edges, which should produce a good result I think.

PR #464, opened 2025-09-02
Performs rotation using shear transformation.

clivepaterson:
This re-implements the rotation algorithm, to use rotation by shear (a pixel preserving rotation method)

Implements rotation by shear:
https://en.wikipedia.org/wiki/Shear_mapping#Rotation

Found in this video:
https://www.youtube.com/watch?v=1LCEiVDHJmc

### DIFF
--- a/shared/src/main/java/com/faforever/neroxis/mask/Mask.java
+++ b/shared/src/main/java/com/faforever/neroxis/mask/Mask.java
@@ -278,15 +278,49 @@ public abstract sealed class Mask<T, U extends Mask<T, U>> permits OperationsMas
         }
     }
 
+    /**
+     * Rotates a point (x, y) around the center of a square of size {@code getSize()}
+     * using a three-shear decomposition of the rotation matrix:
+     *
+     * <pre>
+     * R(θ) = ShearX(-tan(θ/2)) * ShearY(sin(θ)) * ShearX(-tan(θ/2))
+     * </pre>
+     *
+     * Coordinates are first translated so the square’s center is the origin,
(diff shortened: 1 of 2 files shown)


clivepaterson:
Hmm so this doesn't actually work 100%:
image

It almost works, but it has very small artifacting alone the symmetry lines.
And I eventually realised it's because the rotation angle is changing for each slice, so that mean's it doesn't hit every pixel in the end. It almost works, but not fully. The resulting maps are still pretty good and totally playable.

Sheikah45:
I updated the get Rotated point in SymmetryUtil. And do you think the artifact is from rounding of the angle?

clivepaterson:
> I updated the get Rotated point in SymmetryUtil. And do you think the artifact is from rounding of the angle?

Well I've had a closer look and yeah it's interesting how it works..
For example this is a 5 team symmetry with this shear rotation:
image
Which look kinda intersting.

For comparision this is the same noise image, but using the reverseLook algorithm:
image

So yeah, the reverse lookup method does seem better.

And here is a look at map generatered with shear rotation, I can see some of the features of this rotation method:
image
But it still does a fairly good job.

Sheikah45:
Well this is still better than what currently exists I think so might be good to put it in

clivepaterson:
Yeah, it works ok I think. Still not as good as the reverse lookup method.
But it's kinda nice how you can just change the getRotatedPoint() functions. Much easier than the reverse lookup to implement.

I let you decide. I would probably merge it.

It can still be refactored out in future for a reverse lookup method, but not sure if either of us will find the time/motivation to have another go at reverse lookup.

PR #465, opened 2025-09-03
Fixes an issue with the visual debugger not being able to display the visual name for a mask.

clivepaterson:
This fix allows the Visual Debugger to display the mask name correctly:
image

### DIFF
--- a/shared/src/main/java/com/faforever/neroxis/mask/Mask.java
+++ b/shared/src/main/java/com/faforever/neroxis/mask/Mask.java
@@ -845,7 +845,6 @@ public abstract sealed class Mask<T, U extends Mask<T, U>> permits OperationsMas
         try {
             U copy = (U) clazz.getDeclaredConstructor(clazz, String.class).newInstance(this, maskName);
             copy.setVisualDebug(isVisualDebug());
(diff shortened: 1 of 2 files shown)


Sheikah45:
One thing with this is that the copy could diverge in development from the original but this would sort them together in the visual debugger which could be misleading

clivepaterson:
Oh yep, I think I saw that too while building the fractal generator.
I was having to call `setVisualName()` on the copy to change it.

But yeah , I found  it useful for me because the list in the visual debugger can get very large sometimes and this also seems to sort the list better.

But if it's not useful, I can just make this change locally when I need it. And you can just close this pr, it's all good.

Sheikah45:
I think it might be helpful to keep the copies distinct. I have tried to play around with ways to still group the copies together but haven't gotten anything I really like yet

clivepaterson:
> I think it might be helpful to keep the copies distinct. I have tried to play around with ways to still group the copies together but haven't gotten anything I really like yet

I see, good point. Yep... so I had a play. And I think I copied the visual name in the wrong par tof the code...
I've tweaked it and now the names come out like this:
image

Maybe this is better, not sure.

PR #474, opened 2025-11-23
Updates to spawnPlacer for LastTerrain's and fixes to symmtry lines.

clivepaterson:
As reported by Ravandel:
mapName = "neroxis_map_generator_snapshot_htauvztrkz3kc_bifq";
image
The `fillCircle()`call in `SpawnPlacer.java:tryPlaceSpawns()` will either draw a symmetry line or a circle.
So the above mapgen is doing this:
Current_After_fillCircle
Current_After_limitToSymmetryRegion

## Summary by CodeRabbit

* **Improvements**
  * Spawn placement now respects symmetry-driven exclusion zones when shaping spawn regions, improving team separation and consistency.
  * Terrain generation uses the selected symmetry to determine mountain/heightmap processing for a narrower set of symmetry types, producing more predictable terrain.
  * Symmetry line drawing expanded and parameterized to include additional symmetry patterns for better visual consistency.

* **Tests**
  * Map-generator tests updated to use larger map size (512).

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/generator/src/main/java/com/faforever/neroxis/generator/terrain/FractalNoiseLastTerrainGenerator.java
+++ b/generator/src/main/java/com/faforever/neroxis/generator/terrain/FractalNoiseLastTerrainGenerator.java
@@ -52,7 +52,7 @@ public class FractalNoiseLastTerrainGenerator extends MultiLevelLastTerrainGener
         waterHeight -= fractalParams.waterHeight();
 
         symmetryLines.setSize(map.getSize() + 1);
-        symmetryLines.drawSymmetryLines();
+        symmetryLines.drawSymmetryLines(symmetrySettings.terrainSymmetry());
     }
 
     @Override
@@ -138,7 +138,7 @@ public class FractalNoiseLastTerrainGenerator extends MultiLevelLastTerrainGener
 
 
         // Blur and add mountains along the line of symmetry
-        if (Set.of(Symmetry.QUAD, Symmetry.DIAG, Symmetry.POINT2, Symmetry.POINT3, Symmetry.POINT4, Symmetry.POINT5,
+        if (Set.of(Symmetry.POINT2, Symmetry.POINT3, Symmetry.POINT4, Symmetry.POINT5,
                    Symmetry.POINT6, Symmetry.POINT7, Symmetry.POINT8, Symmetry.POINT9, Symmetry.POINT10,
                    Symmetry.POINT11, Symmetry.POINT12, Symmetry.POINT13, Symmetry.POINT14, Symmetry.POINT15,
                    Symmetry.POINT16).contains(symmetrySettings.terrainSymmetry())
--- a/generator/src/test/java/com/faforever/neroxis/generator/MapGeneratorTest.java
+++ b/generator/src/test/java/com/faforever/neroxis/generator/MapGeneratorTest.java
@@ -204,7 +204,7 @@ public class MapGeneratorTest {
     @Test
     public void TestInequalityTournamentStyle() throws Exception {
         MapGenerator instance1 = new MapGenerator(true);
-        new CommandLine(instance1).execute("--tournament-style", "--map-size", "256", "--spawn-count", "2");
+        new CommandLine(instance1).execute("--tournament-style", "--map-size", "512", "--spawn-count", "2");
         SCMap map1 = instance1.getMap();
         long generationTime1 = instance1.getGenerationTime();
         long seed1 = instance1.getBasicOptions().getSeed();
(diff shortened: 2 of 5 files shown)


shared/src/main/java/com/faforever/neroxis/map/placement/SpawnPlacer.java
```diff
@@ -81,10 +81,23 @@ public boolean placeSpawns(int spawnCount, BooleanMask spawnMask, float teammate
...
+                              .fillCircle(mapSize / 2f, mapSize / 2f, mapSize / 4f, true);
+        }
+
         BooleanMask spawnMaskCopy = spawnMask.copy();
-        spawnMaskCopy.fillSides(map.getSize() / spawnCount * 3 / 2, false)
-                     .fillCenter(teamSeparation, false)
-                     .fillEdge(map.getSize() / 32, false)
+        spawnMaskCopy.fillSides(mapSize / spawnCount * 3 / 2, false)
+                     .subtract(teamSeparationMask)
+                     .fillEdge(mapSize / 32, false)
```
Sheikah45: it is a rather big difference to change it from fillCenter to fillCircle. Fill center was filling along the symmetry center. So this would have a big impact on non point symmetry separation. I think just the symmetry lines is enough without changing from fill center to fill circle

Sheikah45: The cause could also be an impropely chosen team symmetry

clivepaterson: Hmm yep, so take2 attempt.
Keeping the `fillCenter()`, which yep, makes more sense to keep it.
And just subtract the symmerty lines.

Which has highlighted some bugs in the `drawSymmetryLines()` function actually...
Seems line the QUAD and DIAG symmetries weren't implemented proplerly in `drawSymmetryLines()`, which I have fixed now, noice....
And yeah, now I see why we need to pass the `SymmetryType` as a parameter to the `drawSymmetryLines()`, so I've updated that aswell, noice...

But yep anyway, this seems to fix these egde cases now where the `fillCenter()` isn't pushing the teams apart enough.
So the above `neroxis_map_generator_snapshot_htauvztrkz3kc_bifq` now does this:
image
image
image
image

generator/src/main/java/com/faforever/neroxis/generator/terrain/FractalNoiseLastTerrainGenerator.java
```diff
@@ -138,7 +138,7 @@ protected void setupHeightmapPipeline() {
 
 
         // Blur and add mountains along the line of symmetry
-        if (Set.of(Symmetry.QUAD, Symmetry.DIAG, Symmetry.POINT2, Symmetry.POINT3, Symmetry.POINT4, Symmetry.POINT5,
+        if (Set.of(Symmetry.POINT2, Symmetry.POINT3, Symmetry.POINT4, Symmetry.POINT5,
```
clivepaterson, line 141: So the QUAD and the DIAG symmetries are both Mirror symmetries, so yep they don't need this "Line of Doom" fixing thing...

I realised this while I was fixing the `drawSymmetryLines()` function.

generator/src/test/java/com/faforever/neroxis/generator/MapGeneratorTest.java
```diff
@@ -204,15 +204,15 @@ public void TestEqualityTournamentStyle() throws IOException {
     @Test
     public void TestInequalityTournamentStyle() throws Exception {
         MapGenerator instance1 = new MapGenerator(true);
-        new CommandLine(instance1).execute("--tournament-style", "--map-size", "256", "--spawn-count", "2");
+        new CommandLine(instance1).execute("--tournament-style", "--map-size", "512", "--spawn-count", "2");
```
clivepaterson, line 207: I ran into some intermittent test failures.
The `assertNotEquals(map1.getHydros(), map2.getHydros());` was failing when both `instance1` and `instance1` generate zero hydro's. Which it seems can happen on 256 maps (5k maps).

I'm not sure what change has caused this, or maybe it's always been the case.
I put the Repeat of 250 on this test:
image

And it fails occassionally. It seems that a bunch of different Styles can have zero hydros on a 256 map:
image
image
image
image

Maybe the SpawnLast has increased the chances of 256 maps having zero hydros? dunno....
So anyway, I increase the mapsize for the Inequality test to 512 (10K). (This should always generate 2 maps with hydros) well maybe 99.99%.

I also generated a bunch of 5K maps, trying to find ones that don't have hydros... Which was hard to find..
And I eventually generated one: (They seem rare)
image

Sheikah45 (approved):

PR #477, opened 2025-12-15
Setons mapgen

clivepaterson:
The Fractal Setons generation....
A lot of work and testing, but I think, it works.

So here is a Collage of 20 maps, 4x4 20k......
_Screenshot removed to speed up Github interface_

Update 22/12/2035:
I did some updates to the Land Bridge algorithm. It was a circle.
Now it's a Square, or Diamond depending on map symmetry.
I've also adjust the water algorithm so that the Land Bridge area is much more likely to be land. (However mounatins are more likely on the Land Bridge now I think).
Anyway here is another 20 maps of 4v4 20k with the updates:
_Screenshot removed to speed up Github interface_

Update 11/01/2026:
Some changes have been made.
The Fractal Navy has copied the Setons water style now, and the water height calculation has been fixed. The fractal water height is now relative to the Biome water height.
And the Islands in the Setons generation are a little higher now, so make the island less vulnerable to T1 navy.
Some other fixes aswell..
This is a  random sample of 4v4 20k maps with all these updates.
collage-20260111-c1

## Summary by CodeRabbit

* **New Features**
  * Added Setonish map style with bridge-based islands and tuned layouts (512/1024, 2 teams).
  * New selectable fractal water masks for varied coastlines and island shapes.
  * Geometry-based drawing utilities (vertex triangle/quad fills) for richer map shaping.

* **Improvements**
  * Spawn spacing adapts to teammate count and map size.
  * Resource placement avoids mex dead zones to reduce blocking.
  * Smoother terrain and heightmap pipeline for more natural mountains and water edges.

* **Removals**
  * Removed Fractal Upside Down map style.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelLastTerrainGenerator.java
```diff
@@ -136,20 +129,56 @@ private void addWaterAreasToNoiseMap(int mapSize) {
...
+
+                switch (symmetrySettings.teamSymmetry()) {
+                    case DIAG, XZ -> {
+                        waterArea.fillRect(0, 0, (mapSize / 2) - padding, (mapSize / 2) - padding, true);
+                        waterArea.fillRect((mapSize / 2) + padding, (mapSize / 2) + padding, mapSize, mapSize, true);
+                    }
+                    case ZX -> {
+                        waterArea.fillRect((mapSize / 2) + padding, 0, (mapSize / 2) + padding, (mapSize / 2) - padding, true);
+                        waterArea.fillRect(0,  (mapSize / 2) + padding, (mapSize / 2) - padding, mapSize, true);
+                    }
```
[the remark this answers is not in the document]
clivepaterson: The fillRect function is defined as:
public U fillRect(int x, int y, int width, int height, T value)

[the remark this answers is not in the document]
clivepaterson: The fillRect function is defined as:
public U fillRect(int x, int y, int width, int height, T value)

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelLastTerrainGenerator.java
```diff
@@ -136,20 +129,56 @@ private void addWaterAreasToNoiseMap(int mapSize) {
...
                 waterAreaBlur.blur(mapSize / 8);
-                break;
+            }
+            case FractalWaterMasks.SETONS -> {
+                int padding = mapSize / 64;
+
+                switch (symmetrySettings.teamSymmetry()) {
+                    case DIAG, XZ -> {
+                        waterArea.fillRect(0, 0, (mapSize / 2) - padding, (mapSize / 2) - padding, true);
+                        waterArea.fillRect((mapSize / 2) + padding, (mapSize / 2) + padding, (mapSize / 2) - padding, (mapSize / 2) - padding, true);
```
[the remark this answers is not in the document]
clivepaterson: This code has changed to:
waterArea.fillRect((mapSize / 2) + padding, 0, rectangleWidthAndHeight, rectangleWidthAndHeight, true);
waterArea.fillRect(0,  (mapSize / 2) + padding, rectangleWidthAndHeight, rectangleWidthAndHeight, true);

please re-check.

clivepaterson:
@coderabbitai, can you scan the changes in this PR for any instances where StrictMath should be used?

shared/src/main/java/com/faforever/neroxis/mask/BooleanMask.java
```diff
@@ -814,9 +814,9 @@ public BooleanMask drawSymmetryLines(Symmetry symmetry) {
                 }
             } else {
                 switch (symmetry) {
-                    case Z, POINT2, POINT4, POINT6, POINT8, POINT10, POINT12, POINT14, POINT16 ->
+                    case X, POINT2, POINT4, POINT6, POINT8, POINT10, POINT12, POINT14, POINT16 ->
                             drawLine(0, halfY, mapSize, halfY);
-                    case X -> drawLine(halfX, 0, halfX, mapSize);
+                    case Z -> drawLine(halfX, 0, halfX, mapSize);
```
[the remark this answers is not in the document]
clivepaterson: A symmetry of X means the top half of the map is a reflection of the bottom half of the map.
A symmetry of Z means the left mirrors the right.

clivepaterson:
@coderabbitai
shared/src/main/java/com/faforever/neroxis/mask/BooleanMask.java:766-767
shared/src/main/java/com/faforever/neroxis/mask/Mask.java:1113,1114,1174,1175
This uses StrictMath now

Learning: In the Neroxis Map Generator codebase: StrictMath should always be used instead of the Math.

generator/src/main/java/com/faforever/neroxis/generator/style/SetonishStyleGenerator.java
```diff
@@ -0,0 +1,65 @@
...
+import com.faforever.neroxis.map.SymmetrySettings;
+
+public class SetonishStyleGenerator extends StyleGenerator {
+    @Override
+    public ParameterConstraints getParameterConstraints() {
+        return ParameterConstraints.builder()
+                                   .mapSizes(512, 1024)
+                                   .numTeams(2, 2)
+                                   .build();
+    }
```
Sheikah45, line 26: Hmm I wonder if we should bring back symmetry constraints to the parameter constraints for this, or move the symmetry settings generation into the style generator

generator/src/main/java/com/faforever/neroxis/generator/style/StyleGenerator.java
```diff
@@ -56,7 +56,7 @@ public abstract class StyleGenerator implements HasParameterConstraints {
     private PropGenerator propGenerator;
     private DecalGenerator decalGenerator;
     private SCMap map;
-    private Random random;
+    protected Random random;
```
Sheikah45: Why was the random made protected? This is a bit risky and should be avoided due to determnism

generator/src/main/java/com/faforever/neroxis/generator/terrain/SetonishLastTerrainGenerator.java
```diff
@@ -0,0 +1,144 @@
...
+            fractalParams = new FractalParams(
+                    14, FractalWaterMasks.SETONS, 2, 1.5f, 5, 2, 8, 50,
+                    List.of(
+                            new FractalFlattenParams(0f, 0.5f, 0, 8, 0.1f, 0, false, false, 4),
+                            new FractalFlattenParams(0.5f, 3f, 8, 16, 4f, 0, true, false, 4),
+                            new FractalFlattenParams(3f, 27, 16, 18, 2f, 1, false, true, 4),
+                            new FractalFlattenParams(27, 50, 18, 24, 0.5f, 1,  false, false, 4)
+                    )
+            );
+        }
```
Sheikah45: We could just add a size parameter constraint of 512. Then if someone wants to make a custom setonish < 512 well then buyer beware

generator/src/main/java/com/faforever/neroxis/generator/terrain/SetonishLastTerrainGenerator.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    protected void setupMountainHeightmapPipeline() {
+        // Draw some mountains specially implemented for Setons
+        rawMountains.setSize(map.getSize() + 1);
+        String brushName = Brushes.GENERATOR_BRUSHES.get(random.nextInt(Brushes.GENERATOR_BRUSHES.size()));
+
+        BooleanMask avoidMountainMask = waterArea.copy().setSize(map.getSize()).deflate(50).setSize(map.getSize()+1);
+        float densityMultiplier = 1f / (1024f / map.getSize());
+
+        if (fractalParams.fractalFlattenParams().size() >= 4) {
```
Sheikah45: Is this if statement necessary it seems like there are always 4

generator/src/main/java/com/faforever/neroxis/generator/terrain/SetonishLastTerrainGenerator.java
```diff
@@ -0,0 +1,144 @@
...
+                 .subtract(waterArea) // For Setons, subtract the water area to prevent spawning on the island
+                 .fillCenter(map.getSize() / 3, false)
+                 .deflate(fractalParams.spawnMaskDeflate());
+
+
+    }
+
+    protected void setupPassablePipeline() {
+        super.setupPassablePipeline();
+        passableLand.subtract(landBridgeBrush);
```
Sheikah45: What is the purpose of removing the landBridgeBrush from passableLand?

The passable masks are assumed to be based on the actual terrain which is why the method was final. That fact is then later used for some of the texture generation for readability.

generator/src/main/java/com/faforever/neroxis/generator/MapGenerator.java
```diff
@@ -222,6 +222,7 @@ public Integer call() throws Exception {
 
     void populateGeneratorParametersAndName() {
         GeneratorParameters.GeneratorParametersBuilder generatorParametersBuilder = GeneratorParameters.builder();
+        //mapName = "neroxis_map_generator_snapshot_4frqi6jkwjs76_ayiaedy";
```
Sheikah45: Can this be removed?

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
```diff
@@ -1019,6 +1019,25 @@ public FloatMask dividePrimitiveWithSymmetry(SymmetryType symmetryType, ToFloatB
...
+        assertCompatibleMask(area);
+        return enqueue(dependencies -> {
+            BooleanMask placement = (BooleanMask) dependencies.get(0);
+            apply((x, y) -> {
+                if (placement.getPrimitive(x, y) && get(x,y) < minValue) {
+                    set(x, y, minValue);
+                }
+            });
+        }, area);
+    }
```
Sheikah45: I think this is just a duplication of the existing clampMin function there is also clampMax in case I got it backwards

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
```diff
@@ -1067,6 +1067,120 @@ protected U fillCoordinates(Collection<Vector2> coordinates, T value) {
...
+    public U fillTriangle(int x1, int y1,
+                             int x2, int y2,
+                             int x3, int y3,
+                             T value) {
+
+        return enqueue(() -> {
+            // Copy parameters to local working variables
+            int ax = x1, ay = y1;
+            int bx = x2, by = y2;
+            int cx = x3, cy = y3;
```
Sheikah45: Why are we copying them the variables are effectively final so should work directly in the lambda?

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
```diff
@@ -1067,6 +1067,120 @@ protected U fillCoordinates(Collection<Vector2> coordinates, T value) {
...
+                int t = bx; bx = cx; cx = t;
+                t = by; by = cy; cy = t;
+            }
+            if (ay > by) {
+                int t = ax; ax = bx; bx = t;
+                t = ay; ay = by; by = t;
+            }
+
+            // Handle degenerate case: all vertices on same Y
+            if (ay == cy) {
```
Sheikah45: It seems like this could be part of the if chain below

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
```diff
@@ -1067,6 +1067,120 @@ protected U fillCoordinates(Collection<Vector2> coordinates, T value) {
...
+                t = ay; ay = by; by = t;
+            }
+            if (by > cy) {
+                int t = bx; bx = cx; cx = t;
+                t = by; by = cy; cy = t;
+            }
+            if (ay > by) {
+                int t = ax; ax = bx; bx = t;
+                t = ay; ay = by; by = t;
+            }
```
Sheikah45: This could be easier perhaps if we create some int vertex object and just put those in a list and sorted by the y.
The one time object allocation is not a big deal.

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
```diff
@@ -1067,6 +1067,120 @@ protected U fillCoordinates(Collection<Vector2> coordinates, T value) {
...
+
+            // Handle degenerate case: all vertices on same Y
+            if (ay == cy) {
+                // Degenerate triangle (horizontal line or point) - just draw a line
+                int minX = StrictMath.min(ax, StrictMath.min(bx, cx));
```
Sheikah45: A little more readable perhaps. Same below
```suggestion
                int minX = IntStream.of(ax, bx, cx).min().orElseThrow();
```

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
```diff
@@ -1067,6 +1067,120 @@ protected U fillCoordinates(Collection<Vector2> coordinates, T value) {
...
+    private void drawScanline(int xStart, int xEnd, int y, T value) {
+        int size = getSize();
+        int start = StrictMath.min(xStart, xEnd);
+        int end = StrictMath.max(xStart, xEnd);
+
+        for (int x = start; x <= end; x++) {
+            if (inBounds(x, y, size)) {
+                set(x, y, value);
+            }
+        }
```
Sheikah45, line 1248: Rather than bound checking each point we should just bound check the start and end and adjust them as clamp them as appropriate

Sheikah45 (changes requested):
I am not sure how finished you consider this but I took a look at some of the functions and have some suggestions for changes.

For the symmetry manipulation I wonder if it would be worth it to move the symmetry settings generation down into the style generators. This way styles could have more control and enable more specificity like we are getting into here. The only thing the end user can specify is the terrain symmetry. But that might be a rework larger than just a simple map type.

generator/src/main/java/com/faforever/neroxis/generator/style/StyleGenerator.java
clivepaterson: Oh yep, it was because I was using `random` in the SetonsStyleGenerator when selecting random symmetries...
However that code has changed, so not needed.
I have made is private again.

generator/src/main/java/com/faforever/neroxis/generator/terrain/SetonishLastTerrainGenerator.java
clivepaterson: Yep, I had the < 512 fallback because the unit testing was failing.
A lot of the unit tests use 256 size and they ignore ParameterConstaints.

However with all the recent changes it seems to actually pass units tests. Noice.
So I have removed the < 512 fallback stuff.

generator/src/main/java/com/faforever/neroxis/generator/terrain/SetonishLastTerrainGenerator.java
clivepaterson: Not really needed, I only added it for the CodeRabbit AI review.
I've removed it...

generator/src/main/java/com/faforever/neroxis/generator/terrain/SetonishLastTerrainGenerator.java
clivepaterson: This was the easist way to prevent Mexes from being placed on the Land Bridge.
I'm not sure if there's a better way to tell the `MexPlacer` that is can't place mexes on the land bridge.

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
clivepaterson: Intellij and the Java compiler was complaining that these variables aren't `effectively final` .
It doesn't like that the variables are changing when they are sorted below.

I guess it's a complexity with the Stack of the main function and the Lambda function.
I'm not sure of the details, but I guess the Lambda function can't alter the stack variables of the outer function, but it can read them.

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
clivepaterson: Let lol, true,  I have updated this... (put it in the main if statement)

Funny story..
The entire `fillTriangle()` method and the `drawScanline()` etc, was all created by ChatGPT. And it just worked.
I think the AI used this articles algorithm amybe: https://www.sunshine2k.de/coding/java/TriangleRasterization/TriangleRasterization.html

Anyway it worked so I focused on other parts of the Setons mapgen code.

And then CodeRabbit reviewed ChatGPT's drawTriangle implementation.
And Code rabbit suggested this code. So I added it for the AI agent..
CodeRabbit likes to pick up of these kinda of edge cases. (It also like to find potential Divide by Zero cases)

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
clivepaterson: Ok, I've put the bounds checking outside the loop now.

Interestingly I've been watch a bunch of the "Advent of Compiler Optimisation" by Matt Godbolt:
https://www.youtube.com/watch?v=j-BwR-Cw0Gk&list=PL2HVqYf7If8cY4wLk7JUQ2f0JXY_xMQm2

And the C++ compliler will actually re-write this function in a very similar way that I've just done...
But yeah, Java it a different philosophy, they wanted to solve the problem of C++ code behaving different on different machines (So I guess that why they leave optimisation to the JVM instead)

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
clivepaterson: I've updated this, it's an interesting technique... pretty neet.
I was gonna say "What about StrictMath? Isn't that important?" but then I realized that this is integers so it's not needed.

generator/src/main/java/com/faforever/neroxis/generator/style/SetonishStyleGenerator.java
clivepaterson: Yep, lol, admittedly I have been avoiding this...
And just focusd on making Nuggets happy with the MapGen (Which is pretty tough)

And the Setons Symmetries rules are wierd, I've done like 3 or 4 variations to the symmetry selection to finally arrive at this algorihtm.

I admit I don't entirely know how to write this properly into the Mapgen code yet.
Mainly because I think it needs to be a function for the constraint.

Sheikah45: Yeah thinking about it I might actually consider moving it down and just having the terrain symmetry be part of the constraint but that is a bigger project than this.

generator/src/main/java/com/faforever/neroxis/generator/terrain/SetonishLastTerrainGenerator.java
Sheikah45: It might be better to maybe add some resourceDeadZone mask to the terrain generator or something so that it can communicate it that way

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
Sheikah45: oh sorry yeah I totally overlooked that you were reassigning the variables lower down. That makes sense then.

generator/src/main/java/com/faforever/neroxis/generator/terrain/SetonishLastTerrainGenerator.java
Sheikah45: Yeah also looking at it seems like most resource generators are 1 to 1 with a terrain generator. I wonder if it is just that coupled really so it would make sense to combine them. But that is another different project

clivepaterson: Ok, I have done this now. I called it `mexDeadZone` because it only limits mex placements, and reclaim can still be placed in this area.

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelLastTerrainGenerator.java
```diff
@@ -116,18 +116,15 @@ protected void landSetup() {
 
     private void addWaterAreasToNoiseMap(int mapSize) {
         // For water only, we can influence the likelihood of water to occur for different areas of the map
-        waterArea.setSize(mapSize + 1);
-
         switch (waterMask) {
```
Sheikah45: Not now but for the future we will have to watch the complexity of this and maybe split it out into other generaators

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
Sheikah45: Bump on this

Sheikah45 (changes requested):

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
clivepaterson: Oh yep, It does look a lot like 16 bit assembly with all the ax, bx, cx....
I have created a Vertex class, and used a list.

shared/src/main/java/com/faforever/neroxis/mask/FloatMask.java
clivepaterson: Oh yep, I looked around for a function that does this, but I couldn't find it....
I have removed this function and used clampMin(), it is identical.

generator/src/main/java/com/faforever/neroxis/generator/terrain/MultiLevelLastTerrainGenerator.java
clivepaterson: Yep, this function is kinda getting big now.
It wasn't instended to be this big, it just kinda grew and got more complicated over time.

I would be more ideal if this code was moved up to the FractalNavyLastTerrrainGenerator and the SetonsLastTerrainGenerator

clivepaterson: Ok I have move this code to the FractalNavyLastTerrrainGenerator and the SetonsLastTerrainGenerator now.

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
```diff
@@ -1067,6 +1069,109 @@ protected U fillCoordinates(Collection<Vector2> coordinates, T value) {
...
+            fillTriangle(List.of(v1, v2, v3), value);
+
+            // Second triangle: (1, 3, 4)
+            fillTriangle(List.of(v1, v3, v4), value);
+        });
+    }
+
+    public U fillTriangle(List<Vertex> vertexList, T value) {
+        if (vertexList.size() != 3)
+            throw new IllegalArgumentException("vertexList size must be 3: vertexList size is " + vertexList.size());
```
Sheikah45: Rather than a list we can just make it take three parameters and create the list in the method.

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
```diff
@@ -1067,6 +1069,109 @@ protected U fillCoordinates(Collection<Vector2> coordinates, T value) {
...
+                               Vertex v4,
+                               T value) {
+
+        return enqueue(() -> {
+            // First triangle: (1, 2, 3)
+            fillTriangle(List.of(v1, v2, v3), value);
+
+            // Second triangle: (1, 3, 4)
+            fillTriangle(List.of(v1, v3, v4), value);
+        });
```
Sheikah45: Unless I am missing something there is an assumption here about the order the vertices as passed to the function to guarantee the triangles don't overlap. We probably don't want to assume that and should make sure the order is as we want in the function when passing to the two triangle functions.

Sheikah45 (changes requested):

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
clivepaterson: Ok yep I have done this, makes more sense.

shared/src/main/java/com/faforever/neroxis/mask/Mask.java
clivepaterson: Yep you are correct actually.
There is also another assumption that only convex quadrilateral's are provided. So this doesn't work for concave quadrilateral's.

I have renamed the method to fillConvexQuadrilateral()
And I've sorted the vertex's around the center of the quadrilateral.
And maintained the use of Java streams for this.

Sheikah45 (approved):

PR #521, opened 2026-09-13
Add adaptive spawn support

n3roxis:
Adds support for the lobby's adaptive slot states (open / closed / closed but spawn mex) to generated maps.

The placers record which mexes and hydros are each spawn's base resources. The generated script removes those markers in OnPopulate for slots that have no player and are not set to "closed - spawn mex", and the scenario gets `AdaptiveMap = true` so the lobby offers that state. Works for both the default and the unexplored script.


Sheikah45:
One thing to consider here is that it is often the case in ladder and custom games that people generate maps with more slots than they intend to fill for the spawns. In custom games this is fine as they can set spawn mex. But this cannot be done in ladder so probably need too talk to the matchmaker team before this is in to understand what impact that would have. Since this defaults all generated maps to adaptive.

